PR #18, opened 2020-12-03
Cleanup code

Bob620:
Hello!

I saw you using various things such as `==` rather than `===` which is a bad thing generally in modern JS.

I also found several spots you were setting variables inside if statements which can be very confusing because it is possible its unintentional. These should be set before the if statement and then checked, preferably.

Largest thing was the `LiveTL/js/background.js`, here I changed it from chained if statements to a switch statement. Though I put a comment because the last if statement was setting the `request.type` variable and I didnt know if that was intentional. May also be a good idea to include a default to the switch.


KentoNishi:
Looks great, thanks for the contribution! Will keep in mind for future updates

Issue #15, opened 2020-12-02
[Request] Add a contributing guideline.


elias-knodel:
**My problem:**
I have no idea which standard to use when writing code, how to write tests etc.

**The solution i would like:**
Adding a CONTRIBUTING.md or / and CODE_OF_CONDUCT.md which describes the workflow etc. This way i can contribute with (hopefully) clean code that meets your requirements.

r2dev2bb8:
Solved with [commit 7fb2c51e769c92b66e328598641db264870b14c9](https://github.com/KentoNishi/LiveTL/commit/7fb2c51e769c92b66e328598641db264870b14c9).

Bob620:
@r2dev2bb8 I would like to have this re-opened. I do not think that the original issue was addressed as I would also like to make potential code changes but do not have a standard to write code to. 

For example, a commonly used standard is:
https://standardjs.com/rules.html

This allows for anyone to write code  that looks the same across all IDEs and not massively change the code style per contributor. Sometimes the actual module is included as a dev-dependancy, though it is not necessary in all projects as long as a specific code style guide is provided.

Bob620:
I also really don't want to auto-format all the code to some random style that I use and mess up the entire visual of the repo

r2dev2bb8:
The standard you provided seems pretty good, although our code is not currently formatted exactly to that. @KentoNishi, are you fine with the standard given by @Bob620? Also, can our repo be autoformatted to fit that style?

KentoNishi:
Sure yeah, my formatter in vscode is only configured to do tabbing and stuff so I can just change it to two spaces. Idk if this is because I write in other languages but I always use four spaces rather than two. Obviously I realize for JS two is more appropriate. Go ahead!

Bob620:
I usually use tabs so that you can set the actual depth to whatever the user wants, its all down to preference with styles, just want to have a consistent style.

KentoNishi:
@Bob620 could you submit a PR with the proposed formatting style?

yoyoyonono:
There's an extention for vscode that automatically formats to those standards: https://marketplace.visualstudio.com/items?itemName=chenxsan.vscode-standardjs

r2dev2bb8:
So I'll change `CONTRIBUTING.md`` to say to use [this](https://standardjs.com/rules.html) but with semicolons.

KentoNishi:
[...]

PR #20, opened 2020-12-03
chore(reformat) change style to standardjs

Bob620:
solve #15


r2dev2bb8 (approved):
Tested in firefox and it works. Looks good to me.

r2dev2bb8:
Can you pull from the latest develop and then autoformat @Bob620?

PR #23, opened 2020-12-03
All Them Badges

yoyoyonono:
Added a badge for the semistandard as well as some others.


r2dev2bb8 (changes requested):
The license badge should be [AGPL](https://www.gnu.org/licenses/agpl-3.0.en.html) instead of GPL.

yoyoyonono:
should be good now

r2dev2bb8 (approved):

Issue #17, opened 2020-12-03
[Bug] Sharing LiveTL with your friends link does nothing


arbrog:
Browser: Chrome  87.0.4280.66
OS: Windows 10
Adblock: disabled

The "sharing LiveTL with your friends" link currently doesn't do anything. Looking at the [href](https://github.com/KentoNishi/LiveTL/blob/00d12eb4f262fa8632ec829949ba955d2ce3ba78/LiveTL/js/frame.js#L405) it's currently pointing to `javascript:void(0);`. I imagine this isn't so much a bug as just something you haven't gotten to yet.

KentoNishi:
Hm, it's supposed to open the device native share menu. I'll look into it when I have time, thanks for the report!

Candygoblen123:
The Navigator.share() method used is only available in Microsoft Edge 81+ and Safari 12.1+.  We would implement a share screen for browsers that do not support Navigator.share() and use the Navigator.canShare() method to determine which implementation should be used.

KentoNishi:
I figured it out. I forgot to put `await` before getFile lmao

PR #24, opened 2020-12-03
LiveTL for Safari

Candygoblen123:
Here's a list of what I did to get this working.

- use the safari web extension converter to convert the Firefox version of LiveTL to Safari (Chrome didn't work for some reason)
- fixed any bugs that I found
- added a `safari` argument to the make file, for easy compiling 

Some notes: 
- only works with Safari 14.0+ 
- comes as an entirely separate .app (but all safari extensions do)
- needs to have full Xcode (and thus, MacOS) to compile, otherwise won't work
- needs to use a developer account to compile for distribution (free should be fine)
- updates should be easy, due to the use of the Firefox codebase
- no iOS support

Screenshots:

Screen Shot 2020-12-02 at 10 27 37 PM

Screen Shot 2020-12-02 at 10 30 40 PM


Candygoblen123:
Here's a pre compiled build, incase you wanted to have a poke:

[LiveTL.app.zip](https://github.com/KentoNishi/LiveTL/files/5633714/LiveTL.app.zip)

r2dev2bb8:
@Candygoblen123 Can you delete the DS_STORE files and add DS_STORE to gitignore?

r2dev2bb8 (changes requested):
After you delete the DS_STORE and add to gitignore, I will merge

Candygoblen123:
It should be good now

Candygoblen123:
Some more notes:
- When chat is open in a LiveTL window, you will not be logged in with your google account unless you disable "Prevent cross-site tracking" under Privacy settings in Safari
- Issue #17 is not present in Safari, and opens the native system sharing menu as expected.  
- The `persistent` flag in `manifest.json` is not supported in safari, and is ignored.  Any effects that it has is not present in this build.  It does not seem to have any major effect on the functionality of the extension, but it should be noted for the future.  
- Distribution of the extension can be done 2 ways: through the App Store or via a direct download from the GitHub repo.  While the App Store is the preferred way, distributing through the App Store comes at a $99 USD cost every year.  If anyone already is a part of the Apple Developer Program, then we can ask them to put it on the App Store for us.  Distributing without going through the App Store is slightly more difficult, because of the lack of auto-update and lack of Apple notarization.  
- No, I don't have enough money right now to enroll in the Apple Developer Program, sorry about that.

r2dev2bb8 (approved):

PR #26, opened 2020-12-03
Improved chat message parsing

Dragon1320:
Improves parse speed (some benchmarks down below). Times shown below are times taken to run benchmark code.

Translation messages - with [lang] block (minority of messages):

- old: ~2.2s
- new: ~100ms

All other messages - without [lang] block (majority of messages - should make a difference here):

- old: ~2.2s
- new: ~200ms

Test code:

```js
// tested on node 14
// some random discord message
const testTrans = "[en] thats why i told you to look into it and was surprised you were continuing on without even checking it xddd";

const messagesPerRun = 1000000;

console.time("label");

for (let i = 0; i < messagesPerRun; i++) {
  parseFunction(testTrans);
}

console.timeEnd("label");
```

Seems to work on both firefox and edge. Not familiar with all the ways translators tag their messages so if i missed anything, lemme know.

### DIFF
--- a/LiveTL/js/lib/filter.js
+++ b/LiveTL/js/lib/filter.js
@@ -1,6 +1,10 @@
+const MAX_LANG_TAG_LEN = 7;
+
 const langTokens = [["[", "]"], ["{", "}"], ["(", ")"], ["|", "|"], ["<", ">"]];
 const startLangTokens = langTokens.flatMap(e => e[0]);
 const tokenMap = Object.fromEntries(langTokens);
+
+const transDelimiters = ["-", ":"];
 
 /**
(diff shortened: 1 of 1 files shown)


Dragon1320:
ahh ye didnt realise there were other ways of denoting translations other than brackets, gonna fix that now

r2dev2bb8:
Initial benchmarks don't show improved speed, I'm writing better benchmarks which simulate the scenarios you provided to get a better reading of the performance .

r2dev2bb8 (approved):
Benchmarks show that this is more than seven times faster than the previous ``parseTranslation``.

PR #30, opened 2020-12-03
Changes to the MacOS build

Candygoblen123:
- Enabled Hardened Runtime
- Set version and build numbers correctly 
- added license and author messages to the about window
- edited some Xcode settings to make it more convent to build


r2dev2bb8 (changes requested):
Can you create a `.gitattributes` with `*.rtf linguist-vendored` to prevent github from categorizing the project as rich text?

Candygoblen123:
should be good now

r2dev2bb8 (approved):

PR #43, opened 2020-12-05
Save preferences

r2dev2bb8:
Now we can save language translator checklist.


r2dev2bb8:
Do not merge yet, there is still an incompatibility.

PR #71, opened 2020-12-12
Fixed safari scrolling in settings menu

Candygoblen123:
Commented out `position: fixed;` in `.modal .modal-content` in `frame.css`. It appeared to have no effect in Both Safari and Firefox, but you should test with Chrome just to be sure no bugs appear.


Candygoblen123:
Also, the .xcuserstate files should be gone now

PR #72, opened 2020-12-13
fixes for v0.2.3

GrumpyBear57:
Fix an issue where duplicate messages were being added to output
Fix an issue with theming and message backgrounds


r2dev2bb8 (approved):
I see no multiple message bugs in Firefox.

Issue #94, opened 2020-12-20
Integration Testing


r2dev2bb8:
**Is your feature request related to a problem? Please describe.**
We have many integration-level bugs.

**Describe the solution you'd like**
The solution is to have integration tests using Selenium or similar software to automatically test various scenarios for common bugs. 

**Describe alternatives you've considered**
We could have human testers, but that is too slow of a feedback loop and human testing can be unreliable. We could still have human testers testing developer builds after they are integration tested.

**Additional context**
n/a

PR #95, opened 2020-12-21
Add integration tests via Selenium

r2dev2bb8:
This adds integration tests to LiveTL, solving #94. The testing can be done via ``make test`` and requires Chrome and Firefox to be installed. Since Chrome with Selenium doesn't support headless extensions, the tests will use headed Chrome. GitHub actions have not been added although if they were to be added, they should only be added for Firefox as Firefox can run extensions with Selenium headless. Only very basic tests are in this and the test cases should be expanded.


r2dev2bb8:
Do not delete this branch, we will be using this branch for testing automated integration tests for pushes.

PR #100, opened 2020-12-23
Safari updates

Candygoblen123:
Safari's back on the menu boys

make sure I didn't break anything before merging, m'kay?


r2dev2bb8 (commented):
It's not catching translations in chilledcow, I'm going to check if the current catches translations there

r2dev2bb8 (changes requested):
Can you merge develop into this so we can see if it works with the current updates?

PR #112, opened 2021-01-01
Automatically set version information in gradle build

r2dev2:
This automatically changes the version in gradle to the version being released and it changes the version code to the number of releases - 31.


r2dev2:
Also, the tests will say failed as I cancelled them as I wasn't modifying files being tested.

PR #145, opened 2021-01-17
Handle YouTube links in Android

arkon:
Adds LiveTL to the list of apps that can open YouTube links when tapping on one.

Supports the same hosts as the official YouTube app.


KentoNishi:
@arkon I resolved some conflicts, can you check if it's correctly merged before I merge it into dev?

KentoNishi:
@arkon my android setup is a bit broken at the moment, can you confirm it works as expected? If so I'll merge

PR #147, opened 2021-01-24
Fix horny detector

arkon:
The previous check also passed if you happened to be subscribed to Marine and her channel was rendered in the sidebar (even if not visually visible).


KentoNishi:
The name of the PR is a certified :bruh: moment

PR #152, opened 2021-01-29
Refactor WebView logic out of MainActivity

arkon:
Attempting to clean it up.

Also increases the minimum Android version to KitKat, because people who are using something older than that aren't worth supporting.


KentoNishi:
This is great but I did change the url being loaded today, so we'll have to look into merging the latest dev with this. Would love if you could go ahead and do that! I only changed a few lines so should be ez

arkon:
This already includes the URL change that you're talking about (I think), but I rebased and fixed an issue with it not loading.

PR #158, opened 2021-02-21
Handle percent signs when decoding title

arkon:
Tested with https://www.youtube.com/watch?v=cpfIxkvtwfI


arkon:
I'm not entirely sure if these files actually get stitched together the way I think they do, so someone pls actually verify that this change is correct.

KentoNishi (approved):

PR #163, opened 2021-03-06
Update all repo and GitHub Pages references

arkon:


GrumpyBear57:
I think a lot of those links can probably also be changed to refer to https://livetl.app instead of a github pages site.

arkon:
Could do that separately. 🤷‍♂️

KentoNishi (approved):

PR #192, opened 2021-04-28
Fix language filters

ChrRubin:
Settings was saving `lang.lang` while `languageNameCode` stored `lang.name`, causing TypeError for language filters other than English, since both values are the same for English.

Also fixes capitalization on `Bahasa Indonesia`.


r2dev2 (approved):
LGTM

PR #194, opened 2021-04-29
Redraw settings slider on resize

ChrRubin:
Fixes visual bug where settings slider stays at the same position/width when the panel/window is resized.

Based on https://github.com/TheComputerM/svelte-materialify/blob/09eb0efb34941565f409757764be1c85519ffced/packages/svelte-materialify/src/components/Tabs/Tabs.svelte#L43.

Probably should be implemented upstream, but the lib is deprecating so ¯\\\_(ツ)\_/¯

### DIFF
--- a/src/components/Settings.svelte
+++ b/src/components/Settings.svelte
@@ -1,5 +1,5 @@
 <script>
-  import { beforeUpdate, afterUpdate } from 'svelte';
+  import { beforeUpdate, afterUpdate, onMount, onDestroy } from 'svelte';
   import { Tabs, Tab, TabContent, MaterialApp } from 'svelte-materialify/src';
   import UISettings from './settings/UISettings.svelte';
   import FilterSettings from './settings/FilterSettings.svelte';
@@ -44,7 +44,12 @@
     }
     callRedrawSlider = false;
   });
-  window.onresize = redrawSlider;
+  onMount(() => {
+    window.onresize = redrawSlider;
+  });
+  onDestroy(() => {
+    window.onresize = null;
+  });
 </script>
 
 <MaterialApp theme="dark">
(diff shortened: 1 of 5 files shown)


KentoNishi (approved):
Agrees completely with my autoformatter so ✅

src/components/Settings.svelte
```diff
@@ -1,20 +1,58 @@
...
+      callRedrawSlider = true;
+    }
+  });
+  afterUpdate(() => {
+    if (callRedrawSlider) {
+      redrawSlider()
+    }
+    callRedrawSlider = false;
+  });
+  window.onresize = redrawSlider;
```
r2dev2: Can you change this to add the event listener on mount and remove it when the component is destroyed? I think either using onMount and onDestroy or using the svelte:window component will work.

src/components/Settings.svelte
```diff
@@ -1,20 +1,58 @@
...
   export let isStandalone = false;
+  export let isResizing = false;
 
   const settings = [
     { name: 'Interface', component: UISettings },
     { name: 'Filters', component: FilterSettings }
   ];
+
+  function redrawSlider() {
+    const sliderElement = document.querySelector('.ltl-settings-slider');
```
r2dev2: Instead of ``document.querySelector``, get a reference to the parent div of the tabs component using ``bind:this`` and use ``querySelector`` from that.

r2dev2 (changes requested):

src/components/Settings.svelte
KentoNishi: done

src/components/Settings.svelte
KentoNishi: done

src/components/Settings.svelte
ChrRubin: Done on latest commit

KentoNishi: modified and done in the latest

PR #198, opened 2021-04-30
Include owner message as mod message

ChrRubin:
Similar to #70 before it got lost during the rewrite.

Also separated author name from timestamp and changed colors to be consistent with YTC/HyperChat.
Before:
image
After:
image


KentoNishi (changes requested):
Font color is a little hard to see, is it possible to do a lighter shade of purple even if it's not consistent with ytc?

ChrRubin:
With the updated mod color:
image

KentoNishi (approved):

PR #208, opened 2021-05-11
Add scroll to recent button

ChrRubin:
Adds QOL scroll to recent button to LiveTL output.

Works for both Top and Bottom text directions.


KentoNishi:
what a chad you are

PR #209, opened 2021-05-12
Fix/Add smooth scrolling

ChrRubin:
Also took the liberty of refactoring scrolling logic out of MessageDisplay. Goodbye `messageDisplay.parentElement.parentElement.parentElement`.

Side effect of smooth scrolling is now the "scroll to recent" button only appears/disappears when scrolling ends, not immediately while scrolling.


KentoNishi (approved):

KentoNishi:
oof when it's scrolling too fast, it doesn't go to the bottom properly

KentoNishi:
try it out with `.*` as a regex filter

PR #211, opened 2021-05-13
Keep scrolling till recent

ChrRubin:
Update of #209, previously if a new TL/message is added while auto scrolling, scrolling stops before said new message. 

This fix changes auto scroll to keep scrolling until the most recent message is reached for both message updates and when clicking the "scroll to recent" button.

A known issue with this change is that the user can't interrupt the auto scroll until it stops, and neither `on:wheel` nor `on:mousewheel` wanted to play nice when I tried to use them for user interrupt. Personally I think this is pretty minor, and is only a significant problem if you're trying to scroll while TL messages are updating with speeds akin to having `.*` regex filter on (which no one realistically does).

### DIFF
--- a/src/components/MessageDisplay.svelte
+++ b/src/components/MessageDisplay.svelte
@@ -26,7 +26,6 @@
   export let items = [];
   export let updatePopupActive = false;
 
-  let bottomMsg = null;
   let unsubscribe = null;
   onMount(() => {
     const { cleanUp, store: source } = combineStores(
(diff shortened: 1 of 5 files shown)


KentoNishi:
Scrolling is still very jank. Personally I think you should submit the refactor as a separate PR and I'll merge that while we can continue working on smooth scroll until it works correctly.

https://streamable.com/417v44

KentoNishi:
force push omegalul will review later

ChrRubin:
Too many jank workarounds required to make native smooth scrolling work and look nice, will revisit with CSS animations at a later date.

PR #217, opened 2021-05-25
Add a dismiss button to Welcome message

Hughp135:
**Why**
The welcome message is a bit of an eye-soar and takes up a lot of space. Furthermore it never goes away even as translations come in (which makes it harder to notice translation messages coming in), so I thought it would be nice to give the user the option to dismiss it, without otherwise affecting the experience.

**What's changed**
- Added a dismiss button to the right of the "Welcome" text
- Clicking dismiss button hides the majority of the welcome message, and replaces it with a smaller message saying that translations will show up here.
- The message 'Translations picked up from the chat' is removed once the first translated message is picked up

**What didn't change**
- The welcome message still shows up every time the extension is loaded (dismissing only lasts for the current session).

**Screenshots**
Default state:
image

After clicking X:
image

### DIFF
--- a/src/components/MessageDisplay.svelte
+++ b/src/components/MessageDisplay.svelte
@@ -5,15 +5,17 @@
     onDestroy,
     createEventDispatcher
   } from 'svelte';
-  import { Checkbox, Button } from 'svelte-materialify/src';
+  import { Checkbox, Icon } from 'svelte-materialify/src';
   import { sources, combineStores } from '../js/sources.js';
+  import Minimizer from "./Minimizer.svelte";
   import '../css/splash.css';
-  import { Icon } from 'svelte-materialify/src';
-  import { mdiEyeOffOutline, mdiAccountRemove, mdiClose } from '@mdi/js';
+  import { mdiEyeOffOutline, mdiAccountRemove, mdiChevronDown, mdiChevronUp } from '@mdi/js';
   import {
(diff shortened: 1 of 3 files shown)


KentoNishi:
alignment is a bit off so i'll go adjust it later

KentoNishi:
also i think instead of an X, an arrow would make more sense. will commit to the fork when i have time

Hughp135:
Adjusted line-height of the welcome title, should address the alignment issue (updated ss)

KentoNishi:
nice! i do wanna make it so u can make it remembers if the message is collapsed or not so u don't have to minimize it every time, i'll do that

r2dev2:
Also, can you use a minimize icon since that's what the dismiss button is doing?

Candygoblen123:
@KentoNishi Can we make it so that people have the option of re-enabling the welcome message? Just in case they dismiss it without knowing it'll never show up again

Hughp135:
> Also, can you use a minimize icon since that's what the dismiss button is doing?

A minimize icon would come with the implication that it can be expanded again, though.

r2dev2:
> 
> 
> > Also, can you use a minimize icon since that's what the dismiss button is doing?
> 
> A minimize icon would come with the implication that it can be expanded again, though.

Yea, it would be preferred if it can also be expanded later.

PR #218, opened 2021-05-25
Options panel takes up full height of container

Hughp135:
**Why:**
Options page doesn't take up full height of the container, causing the tab menu to jump up and down when switching between "Interface" and "Filters"
It looks a lot nicer if the options just takes up 100% of the height

**Changed:**
change `max-height: 100%` to `height: 100%` (options component).

**Screenshots**
After changes:
image

Before changes:
image

note: scrolling still works

### DIFF
--- a/src/components/Options.svelte
+++ b/src/components/Options.svelte
@@ -27,7 +27,7 @@
   .app {
     padding: 0px !important;
     justify-content: center;
-    max-height: 100%;
+    height: 100%;
   }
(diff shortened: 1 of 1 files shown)


Candygoblen123 (approved):
yeah what was up with that anyway? Good Job

KentoNishi:
image

r2dev2 (approved):

PR #219, opened 2021-05-25
Remove margin from captions box

Hughp135:
**Why**
The captions container has a 20px margin, which I presume was either for the purpose of positioning the captions away from the edge of the screen, or to increase the draggable area.
The margin can prevent you from being able to click elements underneath it, when it appears like you should be able to.
For example, if the captions are close to the youtube videos' bottom bar, the margin around the captions box prevents you from being able to click on the youtube bottom bar, which is frustrating.

With margin applied, you are unable to interact with the youtube videos bottom bar, even though it looks like you can
image

**Changes**
Remove margin from captions box


KentoNishi:
This is necessary for smooth dragging, so gonna reject. If you can find a reliable solution to the resize and drag lag on firefox, i'll merge that

PR #220, opened 2021-05-26
[WIP] Use custom version of jquery-ui-touch

Hughp135:
**Why**
Drag events don't work on windows tablet (haven't tested on android)

**Changes**
- Use an updated fork of jquery-ui-touch, and implement code to load it properly
- add an alias from jquery.ui to jquery-ui-bundle (required for the AMD module to load)
- removed jquery-ui-touch dependency from package.json
- Fixes dragging /resizing on windows tablets at least. Perhaps fixes other bugs, or breaks things which i'm not aware of.

**Notes**
- have imported the file into the repo, rather than publishing to NPM. arguments: 1) the way we load the module might not be aplicable to other projects 2) it's a small file so not really worth of an npm package
- this PR is open to discussion, wouldn't recommend merging immediately until this is tested on all devices


KentoNishi:
time to review after a week of not touching this

KentoNishi:
IT WORKS BABY

KentoNishi (approved):

PR #226, opened 2021-05-28
Fix URLs in TLs

ChrRubin:
Similar to LiveTL/chat#10, this PR fixes URLs in the LiveTL output.

Before:
image

After:
image

My suggested fix also incidentally allows member emotes to be shown in the LiveTL output. I've included it in the commit, but I can remove it if it's not desired. Example:
image

Messages that only contain emotes will never be shown in the LiveTL output, since they don't contain any text and will fail the trim() check.


KentoNishi:
lesgo

PR #232, opened 2021-06-05
Show favicon on Firefox

ChrRubin:
Firefox doesn't automatically use icons provided in the manifest as a favicon for extension pages the way chromium-based browsers do. This PR explicitly sets the favicon for the welcome page and the main LiveTL page.

Do note that popout mode on Firefox will still not show the favicon, which is a reported Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1657278. The favicon on the main LiveTL page is technically being set on the Popout component for whenever this bug gets fixed.


KentoNishi:
bugzilla buggerfox

PR #242, opened 2021-06-18
Migrate to webpack v5

ChrRubin:
Full migration to webpack v5. Mainly wanted to migrate so webpack will allow optional chaining (`?.`) instead of throwing an exception. All the build/dev commands work as usual from what I can tell.

Footnotes of the migration:
- webpack and all webpack-related plugins/loaders have been updated to their latest available versions.
- `write-file-webpack-plugin` is removed since `webpack-dev-server` now has a `writeToDisk` option. (Also the plugin doesn't work in v5)
- `copy-webpack-plugin` now uses a `patterns` array for matching files to copy, instead of calling the plugin multiple times.
- `svelte-loader-hot` is removed since `svelte-loader` now has built-in HMR support.


KentoNishi (approved):
looks good to me, thanks for doin this

PR #247, opened 2021-06-26
MCHAD integration

r2dev2:
This integrates the mchad translations into LiveTL. It has been tested on archived vods like [mio's phasmophobia](https://www.youtube.com/watch?v=h7F4XJh41uo), but hasn't been tested for live translations. The screenshot renderer has also been unified.


KentoNishi (approved):
Small stuff but otherwise ok

PR #249, opened 2021-06-27
LiveTL API Read-only integration

r2dev2:
This enables users to read translations from the LiveTL API for both livestreams and vods. The integration has not been tested yet and the ability to write to the api from within YouTube's chat will come in the translator mode update.

### DIFF
--- a/src/components/settings/FilterSettings.svelte
+++ b/src/components/settings/FilterSettings.svelte
@@ -5,7 +5,8 @@
     language,
     customFilters,
     channelFilters,
-    enableMchadTLs
+    enableMchadTLs,
+    enableAPITLs
   } from '../../js/store.js';
   import {
     Row,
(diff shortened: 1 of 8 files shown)


KentoNishi:
awaiting ur checkbox

GrumpyBear57:
> Oh, I had it just load the registered list on startup of the program and lookup from the registered list on startup. This also handles the case that a translator may change their display name or other information like verification status.

Suppose this should be fine, I just wanted to make sure that the api wasn't going to get hit on the full endpoint by (potentially) hundreds of clients anytime a new translator submitted something, a lot easier to just serve the single TLr compared to all of them.

GrumpyBear57 (commented):
Looks good other than I'm still not sure how it'll handle a new translator who registered and starts sending TLs after the initial list retrieval.

r2dev2:
Aight, I can't seem to figure out how to launch the translator client, gonna assume it works now and test later.

PR #254, opened 2021-07-06
Major interceptor refactor (LiveTL)

ChrRubin:
Refer to LiveTL/chat#17. Both PRs will need to be merged and the submodule updated for this to work.

LiveTL specific changes:

- ChatEmbed includes the extension ID as the `embed_domain` param to simplify checking whether to inject LTL buttons.
- Removed `bgmessage.js`, `interceptor.js`, `parse-chat.js`, and most of `background.js`, as they are replaced in the chat submodule PR.


src/js/sources.js
```diff
@@ -191,34 +167,57 @@ export function ytcSource(window) {
     runQueue();
   };
 
-  const getTimestamp = (data, message) => data.isReplay
-    ? message.showtime
-    : (Date.now() + message.showtime) / 1000;
-
-  const extractToMsg = data => message => ({
-    timestamp: getTimestamp(data, message), message
+  const extractToMsg = () => (message) => ({
```
r2dev2: Not quite sure why you used a ``() => (message) => {`` instead of a ``(message) => {``. This previously needed to be a factory function because it took in the data object for more context, but you seem to have refactored out the need for the data object.

src/js/sources.js
```diff
@@ -191,34 +167,57 @@ export function ytcSource(window) {
...
+    port.postMessage({
+      type: 'registerClient',
+      frameInfo: frameInfo,
+      getInitialData: false
+    });
+    portRegistered = true;
+  };
+
+  const params = new URLSearchParams(window.location.search);
+  const isPopout = params.get('popout');
```
r2dev2: Please move ``isPopout`` and ``params`` to ``src/js/constants.js``

src/js/content_scripts/injector.js
```diff
@@ -116,76 +130,77 @@ function loaded() {
...
-    } catch(e) {
     }
-  };
-  insertButtons();
-  setInterval(()=>{
-    if (document.querySelector('.livetlButtonWrapper')) return;
-    insertButtons();
-  }, 100);
+    createPopup(
+      chrome.runtime.getURL(`popout.html?${popoutParams.toString()}`)
```
r2dev2: No need to call ``.toString()`` when doing string templates.

src/js/content_scripts/injector.js
```diff
@@ -116,76 +130,77 @@ function loaded() {
...
+    const embeddedParams = constructParams();
+    embeddedParams.set('embedded', true);
+    embeddedParams.set('tabid', frameInfo.tabId);
+    embeddedParams.set('frameid', frameInfo.frameId);
+    document.body.outerHTML = '';
+    const iframe = document.createElement('iframe');
+    iframe.style.width = '100%';
+    iframe.style.height = '100%';
+    iframe.style.position = 'fixed';
+    iframe.src = chrome.runtime.getURL(`watch.html?${embeddedParams.toString()}`);
```
r2dev2: No need to call ``.toString()`` when doing string templates as it is implicit.

src/js/content_scripts/injector.js
```diff
@@ -116,76 +130,77 @@ function loaded() {
...
+  const params = new URLSearchParams(window.location.search);
+  if (params.get('embed_domain') === chrome.runtime.id) {
+    console.debug('Chat is in extension, not injecting LTL buttons.');
+    return;
+  }
+
+  /** Start buttons injections */
+  makeButton('Open LiveTL', () => {
+    window.top.location =
+      chrome.runtime.getURL(`watch.html?${constructParams().toString()}`);
```
r2dev2: Same params comment as elsewhere in the file.

r2dev2 (changes requested):
Just some things to fix after you're done with the refactor

src/js/sources.js
ChrRubin: Should I move all other `new URLSearchParams(window.location.search)` and `params.get()` usages to constants as well? They're used in a bunch of other places.

src/js/content_scripts/injector.js
ChrRubin: Not the biggest fan of implicit type conversions, but sure.

src/js/sources.js
r2dev2: Yea, that would make the code a lot better.

r2dev2 (approved):

KentoNishi:
this will get merged indirectly thru a merge of [the android-hybrid branch](https://github.com/LiveTL/LiveTL/tree/android-hybrid)

PR #260, opened 2021-07-15
End to end tests

r2dev2:
End to end tests for

- Buttons being injected
- Opening embed panel
- Resizing embed panel
- MCHAD VOD translations for embed mode
- YTC VOD translations for embed mode
- Embed panel scrolling through translations and scroll to bottom button

### DIFF
--- a/e2e/ublock.py
+++ b/e2e/ublock.py
@@ -1,0 +1,30 @@
+from pathlib import Path
+import requests
+
+from autoparaselenium import Extension
+
+
+web_drivers = Path.home() / ".web-drivers"
+web_drivers.mkdir(parents=True, exist_ok=True)
+
+
+class installations:
+    class chrome:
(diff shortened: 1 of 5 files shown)


e2e/ublock.py
```diff
@@ -0,0 +1,30 @@
...
+import requests
+
+from autoparaselenium import Extension
+
+
+web_drivers = Path.home() / ".web-drivers"
+web_drivers.mkdir(parents=True, exist_ok=True)
+
+
+class installations:
```
KentoNishi, line 11: aren't you supposed to capitalize class names in python?

r2dev2: I'm abusing classes to allow me to make variables with properties accessible via dot notation. I have essentially just made an object.

```python
installations.chrome.url
```

KentoNishi: buruh moment lmao ok

PR #264, opened 2021-07-19
E2e reduce concurrency

r2dev2:
Reducing concurrency seems to make the tests execute faster and not error as much.


KentoNishi (approved):
image

PR #265, opened 2021-07-19
Translator Mode

r2dev2:
It now works on chrome, doesn't use invisible characters to preserve space, and doesn't delte emojis when not enabled


ChrRubin (commented):
Seems good and working on Firefox. My one complaint would be the settings screen being a bit messy with all the checkboxes.

Maybe we could consider making settings its own page in the future so its not limited to the tiny Popout component (which already scrolls awkwardly with all the stuff in it right now).

r2dev2:
I moved the translator mode settings to a separate tab as translator mode doesn't fit in with the rest of the user interface settings

ChrRubin (commented):
Definitely looks better, but I still think settings should be moved to its own page in the near future. I might look into that once this and the refactor PR gets merged.

Here's what the settings look like on my usual panel size lol:
image

r2dev2:
> Definitely looks better, but I still think settings should be moved to its own page in the near future. I might look into that once this and the refactor PR gets merged.
> 
> Here's what the settings look like on my usual panel size lol:
> image

I think if you go to ``./options.html`` relative to the extension url it will load options in a full page.

ChrRubin:
> I think if you go to `./options.html` relative to the extension url it will load options in a full page.

Yeah but it doesn't look great as it is right now and would need at least some style changes or even a redesign if we were to kick it out of Popout.

KentoNishi:
Personally against kicking it into a new page because that'll make it harder for third party sites that'll use the ltl interface to change settings. Currently working with the holodex guys to mount the popout component in holodex so we can standardize the tl experience

KentoNishi:
im totally fine with just a dialog that spans the full page instead of just the tl panel tho

KentoNishi (changes requested):
Couple things
1. We need to figure out how to make the settings panel look good before merging this. The tabs look WAY too crowded
2. auto-prefix should have a customizable prefix. not everyone uses the same tag, and they're very fond of their own tag (ex. des. p likes his tag a lot)
3. usage for macros are not explained at all, needs explainer text. took me a while to figure out to press `/`

KentoNishi:
4. Also make the macro activator customizable. I would imagine people want to use stuff like periods too

r2dev2:
> Couple things
> 
>     1. We need to figure out how to make the settings panel look good before merging this. The tabs look WAY too crowded
> 
>     2. auto-prefix should have a customizable prefix. not everyone uses the same tag, and they're very fond of their own tag (ex. des. p likes his tag a lot)
> 
>     3. usage for macros are not explained at all, needs explainer text. took me a while to figure out to press `/`

We can make stuff like customizable prefix and customizable macro leader in more prs. This is just base translator mode functionality. Cleaning up settings should also be in another pr.

KentoNishi:
we should change the pr target branch then, i don't want to merge to dev and beta test before we get those issues sorted out

KentoNishi (approved):

r2dev2:
> we should change the pr target branch then, i don't want to merge to dev and beta test before we get those issues sorted out

No, this should be merged into develop so that future changes to develop don't mess up this beefy pr. There's no point to merging this into yet another branch.

Either we merge this into develop and do pull requests into develop to address the issues of cluttered settings and adding further customizability, or we make the pr for adding customizability into this branch and make the pr of clustered settings into develop where we can merge back to here.

KentoNishi:
im against merging to dev cuz dev should be the branch that's ready to beta test. i don't think this pr is ready to beta test yet, it's missing too man obvious things. we can't work on the improved settings without the new chat stuff already there, since without the additional tab i think it looks fine. it's only with this new tab that it looks horrendous

so tldr
i don't like `this --> dev, settings --> dev` cuz then dev is not publishable and if we are notified of a breaking change on yt's end, we can't send an update without sending out an unfinished tl mode

what i prefer is `settings --> this, this --> dev` or `this --> settings, settings --> dev`. ig if u don't want to change the base, i'll sync up `improved-settings` to this pr and submit a pr going into this branch

KentoNishi (changes requested):
needs pr from `improved-settings` before merge

KentoNishi:
image
just realized: u can't delete macros unless u clear the text and reload. i think that's a significant oversight, we better add that before merging

KentoNishi (changes requested):
to-do before merge now:

- [x] add the option to manually set the auto-prefix
- [x] allow the user to change the macro trigger key/string
- [x] allow the user to delete macros

after those 2 are done, i think we can properly beta test and release after that

KentoNishi:
image

> add the option to manually set the auto-prefix

done ✅

KentoNishi:
image

i've made the store and ui, just gotta use the macroTrigger store in translator-mode.js now

KentoNishi:
@r2dev2 regarding this:

> allow the user to delete macros

i think the checkbox and delete button should behave differently. checkbox is to temporarily disable a macro, and X is to permanently remove it. right now, u have to un-check and reload for the entry to disappear

r2dev2:
> @r2dev2 regarding this:
> 
> > allow the user to delete macros
> 
> i think the checkbox and delete button should behave differently. checkbox is to temporarily disable a macro, and X is to permanently remove it. right now, u have to un-check and reload for the entry to disappear

Oh wait, the checkbox is just supposed to disable not remove, hmmmmm.

KentoNishi:
also another issue i noticed: macros will delete spaces before the trigger key. ex. if i type

```
[en] /testmacro
```
it'll do this:
```
[en]macrovalue
```
instead of what i want, which is
```
[en] macrovalue
```

KentoNishi:
wait scratch that, nvm. i forgot that i had local changes lmao

r2dev2:
Can @KentoNishi and @ChrRubin decide whether or not to add `width: 3.5em` to the delete button's style? The differences between it's existence and lack of existence are the spacing between the checkbox and delete button and the highlight shadow of the delete button when hovered on.

KentoNishi:
will "check it out" budum tss

src/components/options/CustomMacro.svelte
```diff
@@ -29,4 +34,11 @@
     <TextField dense bind:value={expansion}>Expansion</TextField>
   </Col>
   <Checkbox bind:checked={enabled} />
+  <!--
+    @Kento, @ChrRubin, can you see whether width should also be set to 3.5em?
+    Idk which looks better
+  -->
+  <Button icon class="red-text" style="height: 3.5em;" on:click={deleteMacro}>
+    <Icon path={mdiDelete} />
+  </Button>
```
ChrRubin: Its probably better to use a flex div to center it vertically than hardcoding a height. Right now it looks just slightly off center and the highlight shadow gets stretched.

KentoNishi (approved):

src/components/options/CustomMacro.svelte
r2dev2: @KentoNishi refactored it to use a flex div

ChrRubin (approved):

KentoNishi:
so now we just wait for bug reports for a bit, if nothing happens then we merge yay

PR #271, opened 2021-07-25
Make translations not come up twice

### LINKED ISSUE #270: Translations showing up twice
Due to the latest mchad integration with LiveTL, translations by mchad translators who are syncing with YouTube chat are shown twice: once as mchad translations and another time as LiveTL translations.

r2dev2:
Fixes #270

### DIFF
--- a/src/js/sources-util.js
+++ b/src/js/sources-util.js
@@ -9,9 +9,9 @@ import { Message } from './types.js';
  */
 export function removeDuplicateMessages(source) {
   let lastText = '';
-  // return derived(source, $msg => {
-  //   if ($msg == null || $msg.text.trim() == lastText.trim()) return;
-  //   lastText = $msg.text;
-  //   return $msg;
-  // });
+  return derived(source, $msg => {
+    if ($msg == null || $msg.text.trim() == lastText.trim()) return;
+    lastText = $msg.text;
+    return $msg;
+  });
 }
--- a/src/js/sources-util.js
+++ b/src/js/sources-util.js
@@ -42,4 +42,8 @@ const isRecentMessage = (mostRecent, latency) => msg =>
 
 /** @type {(msg: Message) => (otherMsg: Message) => Boolean} */
 const isDuplicateOf = msg => otherMsg =>
-  msg.text.trim() === otherMsg.text.trim() && msg.types !== otherMsg.types;
+  removeWhitespace(msg.text) === removeWhitespace(otherMsg.text) &&
+    msg.types !== otherMsg.types;
+
+/** @type {(text: String) => String} */
+const removeWhitespace = text => text.trim().replace(/(\W)\W+/g, '$1');
--- a/src/js/sources-util.js
+++ b/src/js/sources-util.js
@@ -1,0 +1,14 @@
+import { derived, Readable } from 'svelte/store';
+import { Message } from './types.js';
+
+/**
+ * Removes duplicate messages that come from mchad sync with ytc
+ *
+ * @param {Readable<Message>} source
+ * @return {Readable<Message>}
+ */
+export function removeDuplicateMessages(source) {
+  return new derived(source, $msg => {
+    return $msg;
+  });
+}
--- a/src/js/sources-util.js
+++ b/src/js/sources-util.js
@@ -7,7 +7,7 @@ import { Message } from './types.js';
  * @param {Readable<Message>} source
  * @return {Readable<Message>}
  */
-export function removeDuplicateMessage(source) {
+export function removeDuplicateMessages(source) {
   let lastText = '';
   // return derived(source, $msg => {
   //   if ($msg == null || $msg.text.trim() == lastText.trim()) return;
--- a/src/js/sources-util.js
+++ b/src/js/sources-util.js
@@ -1,17 +1,45 @@
 import { derived, Readable } from 'svelte/store';
-import { Message } from './types.js';
+import { Message, Seconds } from './types.js';
 
 /**
  * Removes duplicate messages that come from mchad sync with ytc
  *
  * @param {Readable<Message>} source
+ * @param {Seconds} sourceLatency the amount of lag between ytc and other sources
  * @return {Readable<Message>}
  */
-export function removeDuplicateMessages(source) {
-  let lastText = '';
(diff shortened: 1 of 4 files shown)


ChrRubin:
This does not remove duplicates when there are other TL(s) in between the duplicated messages, which can easily happen with YTC delay:
image

src/js/sources-util.js
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Removes duplicate messages that come from mchad sync with ytc
+ *
+ * @param {Readable<Message>} source
+ * @return {Readable<Message>}
+ */
+export function removeDuplicateMessages(source) {
+  let lastText = '';
+  return derived(source, $msg => {
+    if ($msg == null || $msg.text.trim() == lastText.trim()) return;
```
KentoNishi: this only checks lastText. you have to have a set and check if the text is in the set.

also, you should also check the author name, since you wouldn't want to block duplicates from independent translators.

KentoNishi (changes requested):

src/js/sources-util.js
r2dev2: I only had it check `lastText` since Taishi does `Coco: *explains meme` a lot with other translations in between. Also, checking the name won't work since Taishi is known as `Taishi Ch.` on ytc and `Kamashiro Taishi` on mchad. If two translators do exactly the same text right next to each other, we should also probably not display that.

r2dev2:
> This does not remove duplicates when there are other TL(s) in between the duplicated messages, which can easily happen with YTC delay:
> image

Hmm, I see, do you think the duplicate checking should extend for a time period like the last 10 seconds or should it be just over the entire stream?

KentoNishi:
> I see, do you think the duplicate checking should extend for a time period like the last 10 seconds or should it be just over the entire stream?

maybe the last minute? since ytc can be quite delayed at times

ChrRubin:
> I only had it check `lastText` since Taishi does `Coco: *explains meme` a lot with other translations in between.

> Hmm, I see, do you think the duplicate checking should extend for a time period like the last 10 seconds or should it be just over the entire stream?

Probably best to make it based on a time period, although the timing would need to be fiddled with to make sure its long enough to compensate for YTC delay, but short enough for something like Taishi's case to not be considered as duplicates. Another idea is to check whether the duplicates are from different sources, ie one from YTC one from mchad/api, since that's the scenario where we expect duplicates.

r2dev2:
> > I only had it check `lastText` since Taishi does `Coco: *explains meme` a lot with other translations in between.
> 
> > Hmm, I see, do you think the duplicate checking should extend for a time period like the last 10 seconds or should it be just over the entire stream?
> 
> Probably best to make it based on a time period, although the timing would need to be fiddled with to make sure its long enough to compensate for YTC delay, but short enough for something like Taishi's case to not be considered as duplicates. Another idea is to check whether the duplicates are from different sources, ie one from YTC one from mchad/api, since that's the scenario where we expect duplicates.

True, I think the original algo was to mark messages that were from different sources and < 10 seconds apart to be duplicates. I should probably just go with that plan.

ChrRubin (commented):
Ran it through [Haachama's latest shitpost review](https://www.youtube.com/watch?v=YndAV5Z5sXQ). About 30 minutes in and it seems good.

Except for one outlier:
image

This has nothing to do with your de-duplication code though. The message from mchad had 2 spaces before 'helped', while the one from YTC had only 1:
Source|Image
---|---
mchad|image
YTC|image

I tested it out, and it seems that YT automatically truncates multiple consecutive spaces into 1 (probably to combat spam), so I'm guessing Taishi accidentally typo-ed 2 spaces into the mchad client.

Just thought this was interesting to know and share, I don't think this should be handled by the extension though. If anything, this behavior should be done server-side on mchad or on our api.

r2dev2:
> Ran it through [Haachama's latest shitpost review](https://www.youtube.com/watch?v=YndAV5Z5sXQ). About 30 minutes in and it seems good.
> 
> Except for one outlier:
> image
> 
> This has nothing to do with your de-duplication code though. The message from mchad had 2 spaces before 'helped', while the one from YTC had only 1:
> Source 	Image
> mchad 	image
> YTC 	image
> 
> I tested it out, and it seems that YT automatically truncates multiple consecutive spaces into 1 (probably to combat spam), so I'm guessing Taishi accidentally typo-ed 2 spaces into the mchad client.
> 
> Just thought this was interesting to know and share, I don't think this should be handled by the extension though. If anything, this behavior should be done server-side on mchad or on our api.

Hmm, should we maybe compare after stripping out all whitespace?

KentoNishi:
yea either strip whitespace, or (even tho this is overkill) use a string similarity metric like dice's coefficient:

https://www.npmjs.com/package/string-similarity

src/js/sources-util.js
```diff
@@ -0,0 +1,45 @@
...
+  throw new Error('Invalid timestamp format');
+};
+
+/** @type {(mostRecent: Message, latency: Number) => (msg: Message) => Boolean} */
+const isRecentMessage = (mostRecent, latency) => msg =>
+  messageTime(mostRecent) - messageTime(msg) <= latency;
+
+/** @type {(msg: Message) => (otherMsg: Message) => Boolean} */
+const isDuplicateOf = msg => otherMsg =>
+  msg.text.trim() === otherMsg.text.trim() && msg.types !== otherMsg.types;
```
KentoNishi: trimming isn't enough, the string similarity or string without whitespace must be compared. also consider emotes? (altho it's a corner case)

ChrRubin:
> Hmm, should we maybe compare after stripping out all whitespace?

I honestly think it'll just introduce more edge-case problems since language in general sucks. Personally I still think the whitespace shortening behaviour should be done server-side if anything.

KentoNishi (changes requested):
it's mostly good, except for the thing rubin mentioned

r2dev2:
Okay, so I made it such that when there is multiple whitespace detected, it removes everything but the first whitespace since this seems to be YouTube's behaviour with whitespace.

KentoNishi (approved):
i hope whitespace is the only thing yt touches... looks good as long as yt doesn't do more stupid sht

PR #275, opened 2021-07-28
Language code fix

alexkoala:
Korean = 한국어
Korea = 한국


KentoNishi:
the incorrect codes are intentional. KO and JA are automatically caught because they match the first two letters of the name of the language (KOrean, JApanese). On the other hand tags like KR and JP (which are actually quite popular) would not be caught if not for this intentional substitution

KentoNishi:
we don't use weblate (yet), that's coming "soon"

alexkoala:
Fixed PR. This PR only changes korean language name.

KentoNishi (approved):

PR #278, opened 2021-07-30
Remove mod deleted messages from filtered output 

### LINKED ISSUE #269: Remove mod deleted messages from filtered output
With the new messaging system in place LiveTL now has access to author ban and message deletion actions. Removing them from the filtered output should help out whenever there are chat mods dealing with spammers who misuse TL tags.

ChrRubin:
Adds support to handle YTC moderator message deletion/author ban actions. 

By default it will hide the deleted/banned messages, but there is an option under filter settings to show the placeholder text from YTC instead, or to ignore these actions entirely.

I've also updated svelte-materialify to the current latest (0.3.11) to support dropdowns that show its the selected object's name property instead of its value property:

Old (0.3.6)|New (0.3.11)
---|---
image|image

Closes #269.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "jquery-ui-bundle": "^1.12.1-migrate",
     "reconnecting-eventsource": "^1.1.0",
     "ssri": "^8.0.1",
-    "svelte-materialify": "^0.3.11",
+    "svelte-materialify": "0.3.6",
     "vue": "^2.6.12",
     "vuetify": "^2.4.9",
     "youtube-iframe": "^1.0.3"
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "jquery-ui-bundle": "^1.12.1-migrate",
     "reconnecting-eventsource": "^1.1.0",
     "ssri": "^8.0.1",
-    "svelte-materialify": "^0.3.6",
+    "svelte-materialify": "^0.3.11",
     "vue": "^2.6.12",
     "vuetify": "^2.4.9",
     "youtube-iframe": "^1.0.3"
--- a/src/components/Message.svelte
+++ b/src/components/Message.svelte
@@ -15,6 +15,8 @@
   export let showTimestamp = false;
   export let thin = false;
   export let inanimate = true;
+  export let deleted = false;
+  export let messageArray = [];
 
   const dispatch = createEventDispatcher();
 
@@ -27,12 +29,13 @@
   class="message"
   class:inanimate
   class:thin
+  class:deleted
   style="display: {hidden ? 'none' : 'block'}"
 >
   <!-- For screenshot checkmark -->
   <slot />
 
-  {#each message.messageArray as msg}
+  {#each messageArray as msg}
     {#if msg.type === 'text'}
       <span>{msg.text}</span>
     {:else if msg.type === 'link'}
@@ -108,6 +111,7 @@
   .info {
     font-size: 0.75em;
     color: lightgray;
+    font-style: normal;
   }
 
   .chat-link {
@@ -125,5 +129,10 @@
     background-color: rgba(255, 255, 255, 0.25);
     padding: 2px;
     border-radius: 5px;
+  }
+
+  .deleted {
+    font-style: italic;
+    color: #898888;
   }
 </style>
--- a/src/components/Message.svelte
+++ b/src/components/Message.svelte
@@ -16,9 +16,13 @@
   export let thin = false;
   export let inanimate = true;
   export let deleted = false;
-  export let messageArray = [];
+  export let messageArray = null;
 
   const dispatch = createEventDispatcher();
+
+  if (!messageArray && message) {
+    messageArray = message.messageArray;
+  }
 
   $: moderator = message.types & AuthorType.moderator;
   $: owner = message.types & AuthorType.owner;
--- a/src/components/MessageDisplay.svelte
+++ b/src/components/MessageDisplay.svelte
@@ -24,7 +24,7 @@
 
   $: document.body.style.fontSize = Math.round($livetlFontSize) + 'px';
   export let direction;
-  /** @type {{ text: String, author: String, timestamp: String }[]}*/
+  /** @type {{ text: String, author: String, timestamp: String, authorId: string, messageId: string, hidden: boolean }[]}*/
   export let items = [];
 
   let bottomMsg = null;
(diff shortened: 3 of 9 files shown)


package.json
```diff
@@ -66,7 +66,7 @@
     "jquery-ui-bundle": "^1.12.1-migrate",
     "reconnecting-eventsource": "^1.1.0",
     "ssri": "^8.0.1",
-    "svelte-materialify": "^0.3.6",
+    "svelte-materialify": "^0.3.11",
```
r2dev2: Keep it at the old materialify, the latest materialify breaks something in our project irc. We'll be switching to something more stable like smelte right after this too.

src/components/Message.svelte
```diff
@@ -27,12 +29,13 @@
...
   class:inanimate
   class:thin
+  class:deleted
   style="display: {hidden ? 'none' : 'block'}"
 >
   <!-- For screenshot checkmark -->
   <slot />
 
-  {#each message.messageArray as msg}
+  {#each messageArray as msg}
```
r2dev2, line 42: Was changing this from message.messageArray to messageArray necessary?

src/components/MessageDisplay.svelte
```diff
@@ -15,16 +15,18 @@
...
+    TextDirection,
+    YtcDeleteBehaviour
   } from '../js/constants.js';
 
   import IntroMessage from './IntroMessage.svelte';
 
   $: document.body.style.fontSize = Math.round($livetlFontSize) + 'px';
   export let direction;
-  /** @type {{ text: String, author: String, timestamp: String }[]}*/
+  /** @type {{ text: String, author: String, timestamp: String, authorId: string, messageId: string, hidden: boolean, messageArray: any[], deleted: boolean }[]}*/
```
r2dev2: You could import Message from the types.js and extend the type so you don't need to use `any[]`

src/components/MessageDisplay.svelte
```diff
@@ -41,9 +43,43 @@
...
       items = items;
     });
+    const hideOrReplace = (i, bonkOrDeletion) => {
+      if ($ytcDeleteBehaviour === YtcDeleteBehaviour.HIDE) {
+        items[i].hidden = true;
+      } else if ($ytcDeleteBehaviour === YtcDeleteBehaviour.PLACEHOLDER) {
+        items[i].messageArray = bonkOrDeletion.replacedMessage;
+        items[i].deleted = true;
+      }
+    };
```
r2dev2, line 54: You may need to do `items = items` at the end of the function to trigger svelte's reactivity for arrays.

yarn.lock
```diff
@@ -7113,10 +7113,10 @@ svelte-loader@^3.1.2:
...
     svelte-hmr "^0.14.2"
 
-svelte-materialify@^0.3.6:
-  version "0.3.6"
-  resolved "https://registry.yarnpkg.com/svelte-materialify/-/svelte-materialify-0.3.6.tgz#a2d4c42ddfcca4bd36625b0481c8daa62dae3b4d"
-  integrity sha512-tiENt0PkgIGa2alIAu3pAVUakQwNpWpxiudUDNyVo0igDuvPzXy/39R+wvPdM4HabjYUdsfoghiOsUcpHCLxIg==
+svelte-materialify@^0.3.11:
+  version "0.3.11"
+  resolved "https://registry.yarnpkg.com/svelte-materialify/-/svelte-materialify-0.3.11.tgz#bdd0a73961e9b408bb584bf2d85b2bb39e5424e9"
+  integrity sha512-MLtZtdSOHditOdJOYqlB8s7NxIFRbYKXJzij5s9gz5WVbJFVEqdX2TqOleg4wYeMOujtOCm+/x3RP5YGyzRxIQ==
```
r2dev2: Please revert this back to materialify-0.3.6, stuff is broken with our project and materialify-0.3.11

r2dev2 (changes requested):

KentoNishi (changes requested):
remove the materialify changes, since we will be switching away + newer version causes issues like r2 said

src/components/Message.svelte
ChrRubin: Yup, the reactivity doesn't work if its kept as message.messageArray

src/components/MessageDisplay.svelte
ChrRubin: I tried `items = items`, `items[i] = items[i]` and a lot of other things and it wouldn't trigger reactivity until I made the `message.messageArray` to `messageArray` change

ChrRubin:
Ack, reverting means the dropdown's gonna look ugly unless we make a custom component for it.

If you guys remember what exactly was the problem in 0.3.11, could you guys try if 0.3.8 is good? That's the version that introduced the Select change.

ChrRubin:
If not, I'll probably hold on to this PR until we migrate to smelte or smui.

Or just remove the settings options and make it hide all the time.

r2dev2:
> Ack, reverting means the dropdown's gonna look ugly unless we make a custom component for it.
> 
> If you guys remember what exactly was the problem in 0.3.11, could you guys try if 0.3.8 is good? That's the version that introduced the Select change.

Wait, what's wrong with the dropdowns?

Also, idrc what was the problem but there was a lot of reeing involved and I don't wanna go down that route again. Two possible options are:

* keep the current dropdowns and revert materialify
* wait till smelte migration

r2dev2:
Oh, we could also just not use a dropdown select for that option. We could temporarily use something like the radio button and then use dropdown after smelte.

ChrRubin:
> Wait, what's wrong with the dropdowns?

Selects (dropdowns) in svelte-materialify let you use different values in code and for showing in UI, but before 0.3.8 the text shown outside of the list is still the stored code value, you can see an example in the screenshots on OP. 

But yeah I'll see if I can make do with radios.

src/components/MessageDisplay.svelte
ChrRubin: I'm importing MessageItem instead since `Message` is already used by the svelte component, and if I alias it intellisense will just make all of it `any` since its still just an invalid js import.

ChrRubin:
Changed to radios:

image

KentoNishi (approved):
as an american english speaker, i feel like ur spelling of behaviour is foreign lmao

src/components/MessageDisplay.svelte
r2dev2: Perhaps 
```js
const before = items.slice(0, i);
const after = items.slice(i + 1);
items = [...before, {...items[i], hidden: true}, ...after];
```
will work?

I think svelte assumes immutable data when deciding reactivity.

r2dev2: Ok, I tried both with a [demo project](https://gist.github.com/r2dev2/a01e185cd06e36efc574c2935c72ecdf), and seems like `[...before, newItem, ...after]` is the only one that works.

ChrRubin: This feels like way too many array operations lmao. Also, changing `hidden` already works as is, I'm assuming because it's a prop. That's why I made `messageArray` a prop as well.

src/components/Message.svelte
```diff
@@ -15,6 +15,8 @@
   export let showTimestamp = false;
   export let thin = false;
   export let inanimate = true;
+  export let deleted = false;
+  export let messageArray = [];
```
r2dev2: ig since this is convenient for the reactivity, we should keep it. However, it would be nice if it played well with the existing api. Perhaps maybe just have the default be null and then if it is null, set it to `message.messageArray` as a fallback.

r2dev2 (changes requested):

src/components/Message.svelte
ChrRubin: Done in latest commit

r2dev2 (approved):

PR #280, opened 2021-08-01
Make MCHAD translators individual users

### LINKED ISSUE #256: Make mchad TLers individual users
Each user from the mchad api has a username, so we should make it so they can be banned as individual tls rather than turning off the entire mchad integration. Same goes for the livetl api

r2dev2:
This makes MCHAD translators individual users. It also adds the ability to source translations from multiple mchad rooms.

Fixes #256


src/js/filter.js
```diff
@@ -198,3 +201,41 @@ export function deleteFilter(id) {
...
+      text: parsed.msg
+    });
+    return true;
+  }
+  if (!isOtherLangTranslation(parsed)) {
+    callback(msg);
+    return true;
+  }
+  return false;
+};
```
r2dev2: I should delete this as it isn't used

KentoNishi (approved):
Changed some smol stuff, looks gud to me

src/components/MessageDisplay.svelte
```diff
@@ -100,11 +101,15 @@
   export let selectedItems = [];
 
   const banMessage = item => () => {
-    channelFilters.set(item.authorId, {
-      ...channelFilters.get(item.authorId),
-      name: item.author,
-      blacklist: true,
-    });
+    if (item.types & AuthorType.mchad) {
+      mchadUsers.set(item.author, true);
```
ChrRubin, line 106: `mchadUsers` is undefined.

src/components/settings/FilterSettings.svelte
```diff
@@ -63,6 +64,12 @@
     <CheckOption name="MChad (volunteer translators)" store={enableMchadTLs} />
   </Col>
 </Row>
+{#if $enableMchadTLs}
+  <MultiDropdown
+    name="Blocked users (MChad)"
+    store={mchadUsers}
+  />
+{/if}
```
ChrRubin: May I suggest putting both "blocked users" dropdowns next to each other? Either moving the YT one down or the MChad one up.

src/js/filter.js
```diff
@@ -192,3 +195,18 @@ export function deleteFilter(id) {
 export function cleanupFilters() {
   customFilters.set(customFilters.get().filter(f => f.rule));
 }
+
+const lang = () => languageNameCode[language.get()];
+
+export const isTranslation = parsed => parsed && isLangMatch(parsed.lang, lang()) && parsed.msg;
+const isOtherLangTranslation = parsed =>
+  parsed && !isLangMatch(parsed.lang, lang()) && parsed.msg;
```
ChrRubin: This method is unused?

ChrRubin (changes requested):

src/components/MessageDisplay.svelte
r2dev2: My bad, I think I had set `window.mchadUsers` and then forgot about it when I removed it.

src/js/filter.js
r2dev2: Removed

src/components/settings/FilterSettings.svelte
r2dev2: Done

image

r2dev2:
@KentoNishi can you please merge chat submodule into this, idk how merging submodules works.

KentoNishi:
Gochu

KentoNishi:
@r2dev2 done

src/components/settings/FilterSettings.svelte
```diff
@@ -42,14 +43,29 @@
...
+  </Col>
+  {#if $enableMchadTLs}
+    <Col style="padding-top:0px">
+      <MultiDropdown
+        name="MChad"
+        store={mchadUsers}
+      />
+    </Col>
+  {/if}
+</Row>
```
ChrRubin: If the dropdown is too small (or if the name is too long) you can't remove the blocked user cause you can't click the `X` button:

image

MIght want to make it grow/wrap or separate the dropdowns to 2 rows.

ChrRubin (changes requested):

src/components/settings/FilterSettings.svelte
ChrRubin: Or maybe we can just move the `on:Click` to trigger on the whole item instead of just the `X` button?

r2dev2: > Or maybe we can just move the `on:Click` to trigger on the whole item instead of just the `X` button?

This isn't clear enough. I'm just going to have the focused one expand to take up the full width.

r2dev2: Ok, I did the grow one.

ChrRubin (approved):

KentoNishi:
merge this to beta or dev?

KentoNishi:
dev it is bois

PR #286, opened 2021-08-04
Add viewport meta tags to pages

arkon:
This fixes scaling issues in the Android app.

Somebody should probably test this in the browsers too, since I'm not entirely familiar with the process.

### DIFF
--- a/src/empty.html
+++ b/src/empty.html
@@ -3,6 +3,7 @@
 
 <head>
   <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width,initial-scale=1.0">
   <title></title>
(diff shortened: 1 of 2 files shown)


src/empty.html
```diff
@@ -3,6 +3,7 @@
 
 <head>
   <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width,initial-scale=1.0">
```
KentoNishi, line 6: empty.html is used for rendering all .html pages, so i actually don't think u need to add the meta to svelte:head in other files. can u try it without the tags in svelte files?

KentoNishi (changes requested):

src/empty.html
arkon: Yup, that works.

KentoNishi (approved):

PR #293, opened 2021-08-05
Spotlight translator

### LINKED ISSUE #272: Isolate translators
Button to isolate a single translator temporarily

r2dev2:
Fixes #272


KentoNishi:
holy sht ur speed, i'll review later

r2dev2:
> holy sht ur speed, i'll review later

No, I'm pomu.

KentoNishi (changes requested):
image

while being spotlighted, the caption box flashes even when no new messages from that specific user are being shown.

there also needs to be a button to un-spotlight.

KentoNishi:
image

also switched to mdiPin cuz it makes more sense than that weird circle square icon

ChrRubin (changes requested):
In addition to what Kento suggested, this also has a small bug with deduplication, where for example if Taishi's mchad is spotlighted, messages from his YTC will never be shown even if it is the earlier message (I've seen it happen on VODs).

Or even worse, if for whatever reason people somehow spotlight his YTC, it'll almost never have messages lmao

r2dev2:
I'm going to refactor this entire thing later into having a store which has all the translations and a store which has all the displayed translations. The future refactor will allow us to have better unbanning and unspotlighting.

ChrRubin (changes requested):
There's some strange behaviour when spotlighting an mchad user:
- Messages from other users that are already shown aren't hidden. Future messages from other users do not get shown, which is expected behaviour.
- The latest mchad message will be duplicated.
- The unspotlight button does not appear.

image

ChrRubin:
> I'm going to refactor this entire thing later into having a store which has all the translations and a store which has all the displayed translations. The future refactor will allow us to have better unbanning and unspotlighting.

Can probably do that when we get to the smelte migration. I've already done a few breaking changes to the background messaging pipeline in smelte HC, and I'm considering moving the entire YTC queue logic to the background script so HC and LTL only have to deal with showing/filtering the messages.

r2dev2:
> > I'm going to refactor this entire thing later into having a store which has all the translations and a store which has all the displayed translations. The future refactor will allow us to have better unbanning and unspotlighting.
> 
> Can probably do that when we get to the smelte migration. I've already done a few breaking changes to the background messaging pipeline in smelte HC, and I'm considering moving the entire YTC queue logic to the background script so HC and LTL only have to deal with showing/filtering the messages.

I meant like on a higher level. Like have the ``items`` reactive array be a store that is globally available so we can globally see which translations are displayed and which are hidden. The ``sources`` object would likely not be changed and instead this would use an ``aggregates`` object.

KentoNishi:
image

KentoNishi (approved):

KentoNishi:
i approb

KentoNishi (approved):

ChrRubin (approved):

PR #298, opened 2021-08-11
Keep it snimple snupple

comradekingu:


CONTACT.md
```diff
@@ -0,0 +1,5 @@
+Welcome to the LiveTL community
+
+Grant desired privacy while offering constructive critique.
+Settle disputes within these means.
+Finding yourselves unable, contact the project team at the [Discord server](https://discord.gg/uJrV3tmthg).
```
r2dev2, line 5: I like the simplicity of this, but can you add this as a 
```markdown
## TLDR
```
to ``CODE_OF_CONDUCT.md`` so we still have the regular COC. It's nice to have the big document since it is a standard COC that works well in other free and open source software projects.

r2dev2 (changes requested):

CONTACT.md
comradekingu: @r2dev2 My point is to avoid it since it sometimes creates awful situations with the power it diverts and conflicts it escalates in secrecy. The 2.0 version is worse yet.
The only thing I would add is "answered by ____ the project maintainer(s)." to have a name to go by.
My little text makes no assumptions, and is recited from memory. Something just like it works for Weblate and in other projects, even when removing that very text. Hosted Weblate doesn't have one, and neither does my small group of translators.
I wouldn't have contributed if I had seen it from the Weblate component route I took, but it is listed on this main repo.
Love the effort. One group of translators I am in resorted to doing something similar with closed source, so full marks.

KentoNishi:
i don't really like this tbh. the original coc is not very long and it's much more clear on the specifics of conduct

r2dev2:
We (the core team) had an internal discussion on this and decided to keep the current COC as we haven't run into issues and this COC is an established COC.

comradekingu:
@r2dev2. Well within your power to keep that out of the community, which is exactly what it is. Guess I leave with my principles and hope you remove the nb_NO translations. I didn't pledge allegiance to it, and I am volunteering my free time.

What you want to do for specific conduct (not vague generalizations and caveats to increase reach of authority) is to adopt the 2.0 variant. Do read it, nobody else does.

GrumpyBear57:
I'm honestly really confused as to what you're trying to accomplish here. You submit a PR to change our existing CoC to something entirely meaningless, under the wrong file, and declare that the 2.0 version of what we're using is worse. Then, after we reject your proposal, you suddenly suggest that we adopt that 2.0 version? You'll have to enlighten me as to why you're even suggesting any of this at all, maybe you could provide an actual example of how an "awful situations with the power it diverts and conflicts it escalates in secrecy" could come about, because I have no idea what you mean. 

Our CoC has never caused us any issues, and we have several active contributors, and a healthy amount of one-time contributors, all of whom agreed to and are able to follow the existing CoC without issue. The only time the CoC has ever even been brought up is now, with you obviously having some sort of issue with it. That being the case, I'd advise you to simply *not make contributions*, rather than coming in after you have, and attempting to overturn it without a concrete explanation or reasoning. 

With regards to your nb_NO contributions, I honestly don't see what reason we should have to remove them. You did agree to the CoC, whether you like it or not, by making that contribution (which by the way, isn't even in this repository). The rest of us working on the project are also volunteering our free time, and also have principles that we'd like to uphold. Principles of which you seem to *not* want to uphold, seeing as you proposed to remove them from our CoC. If you're unable or unwilling to comply with them, then quite frankly, move on.

ChrRubin:
Another active contributor here. While I wasn't present during the discussion earlier due to time zones, I've skimmed through a few of your other similar PRs in other repos, especially the one on the Weblate repo, which was quite an... interesting read, to say the least. It feels like you have a personal vendetta against either the CoC, or even the creator of the CoC, going to great lengths finding repos that use it and opening PRs requesting to remove it. 

Most of the points you've made in the Weblate repo are extremely opinionated (philosophical, even). While we're all entitled to our own opinions and principles, rather than just personal interpretations of the wording used, I'd be more than willing to hear concrete examples of how the CoC can negatively impact our contribution environment. If in the end it does seem like the current CoC would cause more harm than good for us, I'm sure we'll look into changing it.

I'm also confused that you seem to be against the "vague generalizations and caveats" of the current CoC, yet you present even more vague statements as a replacement.

comradekingu:
@GrumpyBear57 I put some rationale into writing.

I believe "Grant desired privacy while offering constructive critique.
Settle disputes within these means." is to the extent possible a deescalation of conflict, as it only concerns the parties directly involved, if possible. That isn't a CoC, so I don't think it is wrong to not use the mechanism to remind everyone it is placed.

I make the distinction between "Code of Conduct" and "contact field", or ~"motivational text".
If writing "we are warm people, and we care about each-other" is how to arrive at such a scenario I am all for it, but that is beside the point. Sure it qualifies for "conduct", but my actual gripe is that it always concerns the utilization of power, aside from the above. "Code" seems to necessitate some penal system to enforce itself too.

Some guideline for behavior is also far removed form the question of misrepresenting power, and then we are closer yet to the parts that aren't redundant to law in the Covenant CoC, or similar efforts.

Someone evidently care somewhere in between tagging up one repo with an old version of the covenant CoC,
and leaving everything else down to assumption. Why that is we don't know. No assumptions made.

If I contribute on Weblate and check both concerning repos, you didn't even make it to missing out on _my_ contributions.
I somehow care, so much so that I don't want to misrepresent what it says with my contributions, and if a main repo has it, should it not apply everywhere? Guess not contributing somehow qualifies, and anyone is now free to keep my contributions in any event, (subject to the great license).

If you start asking what you have in terms of a document, that is a lot better than "it is/needs to be a CoC", or "that is the way it is and it works", because the latter is at most nothing.

For consent, it entirely could be be the case that a some project asked contributors to agree to instating a CoC, or to voluntarily
pledge allegiance once in effect, but that would be a first for me. It seems like no exaggeration that
this scenario is _never_ the case(?), and that it _all_ hinges on implication(?)

From this, false security is drawn, and that is a problem for the most vulnerable of those
that even read it, with good intention at that. Some of which are reduced to 
>socio-economic status, nationality, personal appearance, race, religion, or sexual identity and orientation

to push what I think is bad intentions, and flawed logic.

So far I keep asking for salient reasons to have one or do any of it beyond "turning on green lights",
and I imagine that to be why the now outdated version 1.4 enjoys the most use (to use a word other than popularity).
Would be great if there was good reason, and I am all ears. Is it not the point to reach out to people that can't see past logic that doesn't check out?
Better yet, if any of it worked I would even be in favour of good intentions carrying over even without understanding why.

Whereby the dichotomy of this intention diverges from actual content for me, is with things like
> reasonably be considered inappropriate in a professional setting

Corporate values did not and do not a community make, and similarly

> use of sexualized language or imagery and unwelcome sexual attention or advances

is in not only anti-human, but neither has or could work. It is a departure from intent,
and I believe good intent to be the operational basis for most people, yourselves included, as rational actors.

And so on and so forth.

Removing that very text in Weblate was unproblematic, and community increased greatly (however coincidental.)
To the extent of my rational faculties, I was sadly not wrong about the technical baggage and counter-productive nature of
changing the master branch naming scheme from "master" to "main" in Weblate,
or increases in "diversity" coming to fruition as a result there, or anywhere else.

The spending of Debian on diversity programmes I consider counterproductive, and certainly that is
also the hitherto result. I have donated a fair bit of money to libre software projects,
and I don't want it wasted and inclusivity to hinge on whether one is really self-made or a
product of elevating immutable or otherwise irrelevant characteristics to a class of subsidized
individuals, the list of which at the top of the Covenant CoC does exactly the opposite of
normalization, where that was the norm to begin with.

We have the best meritocracy and tools available found anywhere, and enjoy some level of anonymity if
desired, but somehow "meritocracy" is a central problem according to the creator of the Covenant CoC.
I consistently extend my ideals in not caring about that person in any capacity whatsoever.

I certainly would share the supposed sentiment that all good hopes of putting up a text document would come into fruition,
but sadly that is the concept of "unwritten rules", in writing.

To further codify behavior, vaguely, still means anything anti-social enough to warrant mention should somehow
be off limits to anyone anti-social enough to do it, because now it means disobeying a text-file.

This could be made specific, as with the 2.0, and one could go so far in attempting to undo meritocracy
as to replace "project maintainers" with "community leaders".
That is there for anyone to undermine themselves with, in the hopes of what I don't know.
In "no hugs"-FreeBSD this was certainly not as democratic as one could hope.

I think it also warrants mention that the creator of the Covenant CoC has made a non-libre licensing scheme, which I think
is a further attack on libre software. I certainly hope nobody is trying to represent my ideals in not picking that for their projects.

All this in the wake the mob attempting to remove a Drupal lead dev,
RMS from the FSF under false pretenses, or the guy responsible for rejecting the random number generator
scheme in Linux for stating his opinion and reasons for not exaggerating the language and statistics of rape on collage campuses.

There are enough people asking questions now, meaning the game of CoCs is up.
The benevolent dictator's clothes are at best translucent.

arkon:
I'm going to keep this brief to respect everybody's time.

---

1. The code of conduct is meant to formally codify acceptable behaviour for those participating in the community. More concrete wording is more useful for that.
2. The material text of the document is what's important. External controversies associated with its creators is beyond the scope of its intended use here.
3. We appreciate your volunteered work and we want to continue offering those translations to users unassociated with this discussion. We will therefore not be removing existing translations.
4. We will revisit this if it ever becomes an actual issue for the community.

GrumpyBear57:
@arkon's comment wraps this up well, so I'll refrain from picking apart your entire reply, which seems to just be nonsensical blabbering. 

I hope you'll now realize that we are not so gullible as to accept blatant troll contributions, like Weblate has.

PR #300, opened 2021-08-12
Display mode refactor

KentoNishi:
Replaced `isStandalone` and `paramsEmbedded` usages with a `displayMode` store. This would be helpful for conditionals in the upcoming holodex implementation, where we can just add a new `DisplayMode.HOLODEX` enum

### DIFF
--- a/src/components/Popout.svelte
+++ b/src/components/Popout.svelte
@@ -6,7 +6,7 @@
   import { mdiAccountVoiceOff } from '../js/svg.js';
   import Options from './Options.svelte';
   import Wrapper from './Wrapper.svelte';
-  import { TextDirection, paramsVideoTitle, paramsEmbedded, isAndroid } from '../js/constants.js';
+  import { TextDirection, paramsVideoTitle, isAndroid } from '../js/constants.js';
   import { faviconURL, textDirection, screenshotRenderWidth, videoTitle, enableExportButtons, updatePopupActive, enableFullscreenButton, spotlightedTranslator, isResizing } from '../js/store.js';
   import MessageDisplay from './MessageDisplay.svelte';
   import ScreenshotExport from './ScreenshotExport.svelte';
--- a/src/components/Popout.svelte
+++ b/src/components/Popout.svelte
@@ -7,12 +7,11 @@
   import Options from './Options.svelte';
   import Wrapper from './Wrapper.svelte';
   import { TextDirection, paramsVideoTitle, paramsEmbedded, isAndroid } from '../js/constants.js';
-  import { faviconURL, textDirection, screenshotRenderWidth, videoTitle, enableExportButtons, updatePopupActive, enableFullscreenButton, spotlightedTranslator } from '../js/store.js';
+  import { faviconURL, textDirection, screenshotRenderWidth, videoTitle, enableExportButtons, updatePopupActive, enableFullscreenButton, spotlightedTranslator, isResizing } from '../js/store.js';
   import MessageDisplay from './MessageDisplay.svelte';
   import ScreenshotExport from './ScreenshotExport.svelte';
   import Updates from './Updates.svelte';
   import { saveAs } from 'file-saver';
-  import { isResizing } from '../js/store.js';
   let settingsOpen = false;
   $videoTitle = paramsVideoTitle || $videoTitle;
   $: document.title = $videoTitle || 'LiveTL Popout';
--- a/src/components/Popout.svelte
+++ b/src/components/Popout.svelte
@@ -12,8 +12,8 @@
   import ScreenshotExport from './ScreenshotExport.svelte';
   import Updates from './Updates.svelte';
   import { saveAs } from 'file-saver';
+  import { isResizing } from '../js/store.js';
   let settingsOpen = false;
-  export let isResizing = false;
   $videoTitle = paramsVideoTitle || $videoTitle;
   $: document.title = $videoTitle || 'LiveTL Popout';
 
@@ -153,7 +153,7 @@
   <div
     class="settings-button d-flex"
     class:bottom-float={$textDirection === TextDirection.TOP}
-    class:d-none={isResizing}
+    class:d-none={$isResizing}
   >
     {#if isSelecting}
       <h6 class="floating-text">
@@ -267,9 +267,9 @@
       {/if}
     </div>
   </div>
-  <Wrapper {isResizing} on:scroll={updateWrapper} bind:this={wrapper}>
+  <Wrapper on:scroll={updateWrapper} bind:this={wrapper}>
     <div class:d-none={!settingsOpen}>
-      <Options {isResizing} bind:active={settingsOpen} />
+      <Options bind:active={settingsOpen} />
     </div>
     <div class:d-none={settingsOpen}>
       <MessageDisplay
@@ -282,7 +282,7 @@
       />
     </div>
   </Wrapper>
-  {#if !(isResizing || settingsOpen)}
+  {#if !($isResizing || settingsOpen)}
     {#if !isAtRecent}
       <div
         class="recent-button"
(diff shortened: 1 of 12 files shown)


r2dev2 (approved):

src/components/Watch.svelte
```diff
@@ -13,33 +13,34 @@
...
-    paramsEmbedded
+    DisplayMode
   } from '../js/constants.js';
   import ChatEmbed from './ChatEmbed.svelte';
   import Popout from './Popout.svelte';
   import Captions from './Captions.svelte';
 
   document.title = 'LiveTL';
-  let isResizing = false;
+  import { isResizing } from '../js/store.js';
```
r2dev2: Put above the ``document.title`` line or use ``svelte:window`` to set the title.

src/components/Watch.svelte
```diff
@@ -13,33 +13,34 @@
...
   let chatElem, vidElem, ltlElem;
   const resizable = (selector, info) => {
     j(typeof selector == 'string' ? document.querySelector(selector) : selector).resizable(info);
   };
   const convertPxAndPercent = () => {
     [
       [chatElem, $chatSplit == ChatSplit.VERTICAL ? 'width' : 'height', chatSize],
-      [paramsEmbedded ? null : vidElem, $videoSide == VideoSide.TOP ? 'height' : 'width', videoPanelSize]
+      [$displayMode === DisplayMode.FULLPAGE ? vidElem : null, $videoSide == VideoSide.TOP ? 'height' : 'width', videoPanelSize]
```
r2dev2: This is kinda cluttered. You could move the ``videoSide`` ternary to a reactive statement since you use the same expression on line ``113``.

src/components/Wrapper.svelte
```diff
@@ -22,7 +22,7 @@
 
 <div
   style="
-    display: {isResizing
+    display: {$isResizing
```
r2dev2: Can you refactor this to use ``class:d-none`` and ``class:d-grid``.

r2dev2 (changes requested):

src/components/Watch.svelte
r2dev2: nvm you fixed this

src/components/Watch.svelte
```diff
@@ -98,23 +104,22 @@
...
-          style="{($videoSide == VideoSide.TOP ? 'height' : `width`) +
-            `: ${$videoPanelSize}%;`}
-            {$videoSide == VideoSide.LEFT
+          style="{(isTopSide ? 'height' : `width`) + `: ${$videoPanelSize}%;`}
+            {isLeftSide
             ? 'min-width: 10px;'
-            : $videoSide == VideoSide.TOP
+            : isTopSide
             ? 'min-height: 10px'
             : ''}"
```
r2dev2: Please extract this into a reactive statement.

src/components/Watch.svelte
```diff
@@ -98,23 +104,22 @@
...
+    {#if isFullPage && $showCaption}
       <Captions />
     {/if}
     <div
       id="mainUI"
       class="flex 
-        {$videoSide == VideoSide.TOP ? 'horizontal' : 'vertical'} 
-        {$videoSide == VideoSide.RIGHT ? 'reversed' : ''}"
+        {isTopSide ? 'horizontal' : 'vertical'} 
+        {isRightSide ? 'reversed' : ''}"
```
r2dev2: Please use ``class:`` with this. If you don't want to set two class directives, you can just set a default css value and set a class that overrides the default one.

src/components/Watch.svelte
```diff
@@ -127,29 +132,29 @@
       <div
         class="tile autoscale"
         bind:this={chatSideElem}
-        style={$videoSide == VideoSide.TOP ? 'min-width: 100% !important;' : ''}
+        style={isTopSide ? 'min-width: 100% !important;' : ''}
       >
         <div
-          class="flex {$chatSplit == ChatSplit.VERTICAL
+          class="flex {isChatVertical
```
r2dev2: See line 119 comment

src/components/Watch.svelte
```diff
@@ -127,29 +132,29 @@
...
           <div
             class="tile resizable"
-            style="{$chatSplit == ChatSplit.VERTICAL
+            style="{isChatVertical
               ? 'width'
               : 'height'}: {$chatSize}%;
-                min-{$chatSplit == ChatSplit.VERTICAL
+                min-{isChatVertical
               ? 'width'
               : 'height'}: 10px;
```
r2dev2: Just use one ``isChatVertical`` ternary expression to set both ``width|height`` and ``min-width|min-height``.

r2dev2 (changes requested):

KentoNishi:
found my own issues, don't accept this yet

KentoNishi:
@ChrRubin @r2dev2 fixed, plz review

r2dev2 (approved):

src/components/settings/UISettings.svelte
```diff
@@ -37,7 +37,7 @@
     options={Object.keys(TextDirection)}
     store={textDirection}
   />
-  {#if !isStandalone}
+  {#if $displayMode !== DisplayMode.POPOUT}
     <EnumOption
       name="Video side:"
       options={Object.keys(VideoSide)}
```
ChrRubin: Probably better to change to only show on `FULLPAGE`, video side doesn't do anything for embed mode anyway.

src/components/settings/UISettings.svelte
```diff
@@ -59,7 +59,7 @@
   {/if}
 </div>
 <CheckOption name="Show timestamps" store={showTimestamp} />
-{#if !isStandalone}
+{#if $displayMode !== DisplayMode.POPOUT}
   <CheckOption name="Show captions" store={showCaption} />
 {/if}
```
ChrRubin: Same as above, should only show on `FULLPAGE`.

src/components/settings/UISettings.svelte
```diff
@@ -59,7 +59,7 @@
...
 {#if $showCaption}
```
ChrRubin: Extra caption settings will always be shown if captions are enabled regardless of display mode.

src/components/Popout.svelte
```diff
@@ -7,16 +7,14 @@
   import Options from './Options.svelte';
   import Wrapper from './Wrapper.svelte';
   import { TextDirection, paramsVideoTitle, paramsEmbedded, isAndroid } from '../js/constants.js';
```
ChrRubin: Unused import `paramsEmbedded`

src/components/Watch.svelte
```diff
@@ -59,99 +60,91 @@
...
+  $: isTopSide = $videoSide == VideoSide.TOP;
+  $: isLeftSide = $videoSide == VideoSide.LEFT;
+  $: isFullPage = $displayMode === DisplayMode.FULLPAGE;
+  $: isChatVertical = $chatSplit == ChatSplit.VERTICAL;
+
+  $: chatElemStyle = (
+    isChatVertical ? `width: ${$chatSize}%; min-width: `: `height: ${$chatSize}%; min-height: `
+  ) + '10px;';
+  $: videoContainerStyle = (
+    (isTopSide ? 'height' : `width`) + `: ${$videoPanelSize}%; ` +
```
ChrRubin: `width` is in backticks instead of single quotes

ChrRubin (changes requested):

src/js/store.js
```diff
@@ -98,4 +98,6 @@ export const timestamp = writable(0);
 export const faviconURL = writable('/48x48.png');
 export const availableMchadUsers = writable([]);
 export const spotlightedTranslator = writable(null);
+export const displayMode = writable(paramsEmbedded ? DisplayMode.EMBEDDED : DisplayMode.FULLPAGE);
```
r2dev2, line 101: This seems like it should be a constant value though. You only set it once before the app is mounted. Maybe try moving this to constants and getting the page type by setting a ``window`` value before the app mounts or by checking the url.

r2dev2 (changes requested):

src/js/store.js
r2dev2: Nvm, this is better.

r2dev2 (approved):

PR #304, opened 2021-08-14
Suppress spam

### LINKED ISSUE #283: Auto hide spam messages/suppress succesive message within time period
**Checklist (make sure you do all these before you submit a bug report)**
- [x] This is an issue for Hyperchat, not LiveTL. I will go to [the LiveTL repo](https://github.com/LiveTL/LiveTL) to report an issue for LiveTL.
- [x] I have checked for a duplicate or similar issue. I made sure to check closed issues as well. 
- [x] I am not reporting a bug. If I am, I will use the bug report issue template.
- [x] I understand that I am not maintaining this codebase, and my suggestion may be rejected for any reason, including "We don't really feel like it" or "That's a little out of scope for this project" (We accept most suggestions, though.)
- [x] I also understand that even if we do accept your suggestion, it may be removed from the extension at any time, for any reason, without notification. It might not make it to a release version of the extension. 

**Is your feature request related to a problem? Please describe.**
I notice that spam messages usually send multiple messages in quick succession, and sometimes they misuse the prefix for auto-tl. I'm not sure if the deleted messages still shown in the translation section.
image
Screenshot taken from `https://www.youtube.com/watch?v=4p2BaQ9WRzo`

**Describe the solution you'd like**
A way to automatically ignore and hide the messages if the message is sent by the same user within the certain number of time, e.g. 3 successive messages within 500ms? maybe configurable in Filter tab?

**Describe why this solution should be included in Hyperchat**
This will improve the quality-of-live of the chat and translation (for live-tl part).

**Describe alternatives you've considered**
manually add the user/bot to the blocklist, but this is not maintainable especially the bot keep come up with new username.

**If this is suggesting a change to LiveTL's UI, provide documentation for why this change should be made**
N/A

**Discord**
N/A

**Additional context**
N/A

r2dev2:
Fixes #283


src/components/SpamProtection.svelte
```diff
@@ -0,0 +1,30 @@
+<script>
+  import { Subheader } from 'svelte-materialify/src';
```
ChrRubin, line 2: Unused import

src/js/store.js
```diff
@@ -74,7 +74,10 @@ export const
   autoPrefixTag = SS('autoPrefixTag', '[$filterLang]'),
   macroTrigger = SS('macroTrigger', '/'),
   ytcDeleteBehaviour = SS('ytcDeleteBehaviour', YtcDeleteBehaviour.HIDE),
-  autoVertical = SS('autoVertical', true);
+  autoVertical = SS('autoVertical', true),
+  enableSpamProtection = SS('enableSpamProtection', true),
+  spamMsgAmount = SS('spamMsgAmount', 5.9),
+  spamMsgInterval = SS('spamMsgInterval', 0.9);
```
ChrRubin: Is there a reason the default values have decimal places? I thought the slider component stores in pure integers anyway?

src/js/sources-util.js
```diff
@@ -20,9 +20,55 @@ export function removeDuplicateMessages(msgs, sourceLatency=10) {
...
+
+/** @type {(msgs: Message[], amount: Number, time: Seconds) => Boolean} */
+const containsSpam = (msgs, amount, time) => {
+  let beg = 0;
+  const timestamps = msgs.map(messageTime);
+  for (const [i, timestamp] of timestamps.entries()) {
+    // advance beg pointer until it is time seconds behind timestamp
+    while (timestamp - timestamps[beg] > time) beg++;
+    if (i - beg + 1 >= amount) return true;
+  }
```
ChrRubin, line 51: I might just be dumb and need an explanation of how this loop works, but I think line 49 should be a `>=` to include messages that are exactly at the set interval seconds to be counted as spam.

src/components/SpamProtection.svelte
```diff
@@ -0,0 +1,30 @@
...
+    suffix=""
+  />
+  <Slider
+    name="every {interval} second{intervalPlural}"
+    store={spamMsgInterval}
+    min={1}
+    max={98}
+    suffix=""
+  />
+{/if}
```
ChrRubin: I feel like we should show the list of authors that are hidden by the spam protection somewhere so people can tweak the settings in case false positives are hidden.

ChrRubin (changes requested):
Slightly worried about the performance of the many looping and iterations that are being done by this, especially for streams with a lot of TLs, but I guess we can cross that bridge when it starts being a problem.

src/js/store.js
r2dev2: The slider I think stores decimals so I set it to ``.1`` under and did a ``Math.ceil``.

src/js/sources-util.js
r2dev2: No, it should be ``>`` because the end condition that we want is ``timestamp - timestamps[beg] <= time`` and ``<=`` is the opposite of ``>`` so when this loop finishes, the end condition is ``true``.

src/components/SpamProtection.svelte
r2dev2: Fixed in 980d45c.

r2dev2:
> Slightly worried about the performance of the many looping and iterations that are being done by this, especially for streams with a lot of TLs, but I guess we can cross that bridge when it starts being a problem.

Adding a new message has always been ``O(n)`` where ``n`` is the amount of messages already there because svelte does have to iterate through each message to check which message changed. Some of the looping can be optimized away with some clever memoization if it comes to it.

KentoNishi:
I updated the css and took care of alignment issues that were bugging me.

I really dislike using `nbsp` as a hacky way to handle alignment, so if there's a way to avoid it, plz avoid it lol

image

KentoNishi:
I also updated the stores. There's no need to do the sussy wussy floating point shenanigans

r2dev2:
> I also updated the stores. There's no need to do the sussy wussy floating point shenanigans

They get stored as floats when they're modified and also sometimes, sus stuff happens when they're exactly at the tipping point.

ChrRubin:
> They get stored as floats when they're modified and also sometimes, sus stuff happens when they're exactly at the tipping point.

Probably just svelte-materialify or the version we're using, Slider got updated a few times after 0.3.5. Hopefully won't be an issue in smelte

src/js/store.js
r2dev2: @KentoNishi  changed to ``Math.round``.

src/components/SpamProtection.svelte
r2dev2: Done

KentoNishi:
seems to be working

image

KentoNishi:
image

made it look slightly nicer

KentoNishi:
image

KentoNishi:
oops i fcked up hold on

KentoNishi:
there we go, this is a bit better. hopefully we can fix all alignment problems properly using `flexbox` or `grid` instead of using `rem` spacing when we move to smelte

image

KentoNishi:
applied to standard dropdown too

image

KentoNishi (approved):

src/js/sources-aggregate.js
```diff
@@ -78,18 +93,27 @@ export const capturedMessages = readable([], set => {
...
-  [capturedMessages, allBanned, hidden, spotlightedTranslator, ...spamStores];
+const dispTransform =
+  ([$items, $banned, $hidden, $spot, $spamAmt, $spamInt, $whitelisted, $enSpam]) => {
 
-const dispTransform = ([$items, $banned, $hidden, $spot, $spamAmt, $spamInt]) => {
   const attrNotIn = (set, attr) => item => !set.has(item[attr]);
-  const spammers = new Set(getSpamAuthors($items, $spamAmt, $spamInt));
+  const spammers = $enSpam
+      ? getSpamAuthors($items, $spamAmt, $spamInt).filter(([id]) => !$whitelisted(id))
+      : [];
```
ChrRubin, line 109: Kinda on the fence about permanently whitelisting "spammers", since there's no way to un-whitelist them aside from reinstall or manually editing the settings json.

Just worried that there'll be dumb users that "accidentally" whitelist legit spammers lol

ChrRubin (commented):
Seems good. The only part I'm kinda iffy about is that whitelisting spammers is permanent and global, but even that is mostly a user error kind of thing, so I'm okay with merging if we just wanna blame the user whenever that happens lmao

src/js/sources-aggregate.js
r2dev2: You can ban them.

r2dev2:
> Seems good. The only part I'm kinda iffy about is that whitelisting spammers is permanent and global, but even that is mostly a user error kind of thing, so I'm okay with merging if we just wanna blame the user whenever that happens lmao

The user can be banned with the normal ban feature.

ChrRubin (approved):

PR #308, opened 2021-08-14
Duplication check uses epoch ms

### LINKED ISSUE #305: Duplication check stopped working
image

It's probably because the mchad timestamp string was changed to HH:MM AM/PM instead of the previous HH:MM:SS.

Anyway this is a great time to change the duplication check to use epoch milliseconds instead of timestamp strings lol

KentoNishi:
Originally a PR by Rubin, recreated after a failed merge (since tests were not updated)

Fixes #305.

At some point we should probably agree on the same locale to format timestamps for consistency.

Current timestamp locale/format:

Source | Live stream | VOD | Notes
---|---|---|---
YTC | en-GB (HH:MM:SS) | MM:SS and HH:MM:SS | Livestream timestamp used to be based on user locale. VOD timestamp is taken directly from the JSON
Mchad | en-US (HH:MM AM/PM) | HH:MM:SS

(PS: #305 only affects live streams, so I'm waiting for Taishi's next TL session to test this fix there. It works on VODs tho)

### DIFF
--- a/src/__tests__/js/sources-util.js
+++ b/src/__tests__/js/sources-util.js
@@ -4,11 +4,11 @@ import { removeDuplicateMessages } from '../../js/sources-util.js';
 import { AuthorType } from '../../js/constants.js';
 
 
-const message = (text, author, timestamp, types) => ({
+const message = (text, author, timestampMs, types) => ({
   text,
   messageArray: [{ type: 'text', text }],
   author,
-  timestamp,
+  timestampMs,
   id: '9',
   types
 });
@@ -17,9 +17,9 @@ const message = (text, author, timestamp, types) => ({
 describe('message duplication mitigation', () =>{ 
   it('filters duplicate messages from another source within the past 10 seconds', () => {
     const messages = [
-      message('Konichiwassup', 'Taishi', '00:10:00', AuthorType.mchad),
-      message('hello there', 'Taishi', '00:10:12',  AuthorType.mchad),
-      message('hello there', 'Taishi Ch.', '10:15', 0),
+      message('Konichiwassup', 'Taishi', 100000, AuthorType.mchad),
+      message('hello there', 'Taishi', 112000,  AuthorType.mchad),
+      message('hello there', 'Taishi Ch.', 115000, 0),
     ];
     const expectedMessages = [
       messages[0],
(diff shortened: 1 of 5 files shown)


src/js/sources-util.js
```diff
@@ -20,23 +20,9 @@ export function removeDuplicateMessages(msgs, sourceLatency=10) {
...
-    const [minutes, seconds] = timeSegments;
-    return minutes * 60 + seconds;
-  }
-  throw new Error('Invalid timestamp format');
-};
-
 /** @type {(mostRecent: Message, latency: Number) => (msg: Message) => Boolean} */
 const isRecentMessage = (mostRecent, latency) => msg =>
-  messageTime(mostRecent) - messageTime(msg) <= latency;
+  (mostRecent.timestampMs / 1000) - (msg.timestampMs / 1000) <= latency;
```
r2dev2: Can you just change the definition of ``messageTime`` since that is used in removing spam. It is also preferable to have it in a function since I use ``.map`` with it.

r2dev2 (changes requested):

r2dev2:
Also, the tests need to be changed.

src/__tests__/js/sources-util.js
```diff
@@ -48,33 +48,33 @@ describe('message duplication mitigation', () =>{
...
     const messages = [
-      message('*Explains meme', 'Taishi', '00:10:00', AuthorType.mchad),
-      message('*Explains meme', 'Taishi Ch.', '10:05', 0),
-      message('*Explains meme', 'Taishi', '00:10:08', AuthorType.mchad),
+      message('*Explains meme', 'Taishi', 100000, AuthorType.mchad),
+      message('*Explains meme', 'Taishi Ch.', 105000, 0),
+      message('*Explains meme', 'Taishi', 108000, AuthorType.mchad),
     ];
     const expectedMessages = [
       messages[0],
```
ChrRubin, line 80: This test is still failing on local for some reason

r2dev2: Jest is sussy, it works when you reclone the folder.

ChrRubin: Nevermind jest is weird

r2dev2 (approved):

KentoNishi:
omegalul

image

PR #313, opened 2021-08-17
Make api link reactive to filter language

r2dev2:
- change api sse to to change link based on language code
- make api archive link update based on user's language

### DIFF
--- a/src/js/api.js
+++ b/src/js/api.js
@@ -71,8 +71,14 @@ const authorName = (() => {
 const apiLiveLink = (videoId, lang) =>
   url(`/translations/stream?videoId=${videoId}&languageCode=${lang}`);
 
-/** @type {(videoId: String) => String} */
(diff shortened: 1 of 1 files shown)


KentoNishi (approved):
seems like it works properly

GrumpyBear57 (changes requested):
#276 should be used instead, as it will use the api-wrapper package, which fixes several issues with parsing API TLs into local message objects, and moves away from the deprecated `translations/stream/` endpoint. 

It just requires the language code to be implemented to be equal with this PR, but still needs the UI to not display translations at a future timestamp before it's merged.

r2dev2:
> #276 should be used instead, as it will use the api-wrapper package, which fixes several issues with parsing API TLs into local message objects, and moves away from the deprecated `translations/stream/` endpoint.
> 
> It just requires the language code to be implemented to be equal with this PR, but still needs the UI to not display translations at a future timestamp before it's merged.

I'll add the functionality to that branch then.

PR #317, opened 2021-08-18
Fix dictation

### LINKED ISSUE #314: Read aloud mode doesn't work without captions
Right now, read aloud mode is done via ``checkAndSpeak(captionText = text)`` in ``src/components/Captions.svelte``. The ``captionText`` variable should be refactored out into a ``readable`` store where it checks and speaks either in the store or in an imported module.

r2dev2:
Fixes #314

### DIFF
--- a/src/changelogs/6.7.0.svelte
+++ b/src/changelogs/6.7.0.svelte
@@ -17,4 +17,11 @@
       />
     </div>
   </ExpansionPanel>
+  <ExpansionPanel>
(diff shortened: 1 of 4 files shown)


KentoNishi (changes requested):
image

the default caption text is gone

ChrRubin (approved):

KentoNishi (approved):

PR #320, opened 2021-08-19
Improved speech synth

KentoNishi:
Adds support for switching read-aloud mode voices, as well as the speech speed.

The `store` fckery I did in `ReadAloud.svelte` is purely to deal with Materialify being stupid. Voice names usually have spaces, and Materialify can't handle dropdowns when the `value` of a dropdown entry has whitespaces. So the store just uses a stringified integer for the value, converting it back to the correct voice name when the value changes.

The speech synthesis API also only returns valid values for voices and such after the page is fully loaded, so the readable and derived stores exist for that reason. The actual name of the voice is stored in `speechVoiceNameSetting`, and `speechVoiceName` is set to either the stored value or default value. This also prevents displaying nonexistent voices when a user imports settings from a different browser (ex. Edge has dozens more voices than Firefox, and importing the settings json with an Edge voice will make it switch to the default voice).

I tested this on Edge and Firefox, and it worked in both

### DIFF
--- a/src/components/options/Slider.svelte
+++ b/src/components/options/Slider.svelte
@@ -69,7 +69,7 @@
 </script>
 
 <div bind:this={wrapper}>
-  <Slider bind:value {color} {step} thumb={thumb ? true : false}>
+  <Slider bind:value {color} {step} thumb={Boolean(thumb)}>
     <div slot="default">
       {name}{suffix}
     </div>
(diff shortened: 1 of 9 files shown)


ChrRubin (approved):

src/components/options/Slider.svelte
```diff
@@ -69,7 +69,7 @@
 </script>
 
 <div bind:this={wrapper}>
-  <Slider bind:value {color} {step} {thumb}>
+  <Slider bind:value {color} {step} thumb={thumb ? true : false}>
```
r2dev2: You could use ``Boolean(thumb)``

r2dev2 (approved):

PR #334, opened 2021-09-05
Smelte migration

### LINKED ISSUE #221: No way to open stream in LiveTL after using Embedded TLs
If you click on 'Embed TLs', there is no way to open the full LiveTL view without reloading the page. This creates a 'dead end' trap where there is no easy way to change between embedded/ full view unless you refresh the page or hit the browser back button.

The UX from the start should be more intuitive to a first time user, who may not know what 'Embed TLs' and 'Open LiveTL' even do.

image

ChrRubin:
Most of the migration work is done, with only a couple of tasks and bugs we should deal with before we merge into develop:

- [x] ~Font size setting gets overridden by Tailwind classes~
- [x] ~MultiDropdown doesn't have a max width set~
- [x] ~Migrate jest tests~
- [x] ~Migrate e2e tests~
- [x] ~Build/Package scripts~
- ~I couldn't get the build/webserver scripts to behave nicely when setting up smelte, so I went with the lazy route and used webpack-cli. We could just completely replace them if we want, since I've already did the chrome manifest thing inside CopyWebpackPlugin, and version can be set with --env~
- [ ] Test with Android hybrid
- [x] ~Style and design tweaks~
- ~Need to run through and tweak some of the styles even more, e.g. the Update dialog having a pretty jarring shade difference when used with ExpandingCard~


r2dev2:
aya
155 changed files

KentoNishi:
I didn't even know we had that many files to begin with

ChrRubin:
I'll wait for #343 before attempting to merge 6.8, don't want to resolve 2 conflicts instead of 1 lmao

ChrRubin:
(Just taking a note here so I don't forget)

Might be an interesting idea to have embed mode inject the Popout component instead of replacing the whole chat. It'll lose resizability, but sending superchats and whatever YT decides to disable for embedded chats would work for embed mode on the YT page.

KentoNishi (approved):
Android hybrid support can come later, looks good to merge for desktop!

r2dev2 (approved):

KentoNishi:
After so many months, it's finally releasable! Y E S !

Now I know this is kinda cringe right, but I really appreciate you @ChrRubin
:kiaralove:

PR #296, opened 2021-08-06
E2E tests for LiveTL UI

r2dev2:
This has tests for captions. If anyone has any ideas for more tests for the `Open in LiveTL` view, let me know! I plan on adding in tests for the functionality in #294.

PR #337, opened 2021-09-10
player polyfill

r2dev2:
The recent development to embed the player in a YouTube 404 page to bypass embed restrictions broke the existing `window.player` api that our livetl api code, future tests in #296, and future features like `keyboard-shortcuts` depend on. This polyfills that with the caveat that all methods are now `async`.

### DIFF
--- a/src/js/api.js
+++ b/src/js/api.js
@@ -5,6 +5,7 @@ import { APITranslation, Message, ScriptMessage } from './types.js';
 import { AuthorType, paramsIsVOD } from './constants.js';
 import { formatTimestampMillis, sortBy, toJson } from './utils.js';
 import { enableAPITLs, timestamp } from './store.js';
+import { player } from './polyfills/player.js';
 import ReconnectingEventSource from 'reconnecting-eventsource';
 
 export const sseToStream = link => readable(null, set => {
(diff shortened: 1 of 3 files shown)


ChrRubin:
Oh boy I'm REALLY not a fan of how hackish this is, and even CodeQL agrees with me on this one lmao.

Btw the LTL API code you mentioned wouldn't even work in Popout and Embed mode, since there's no player in the context of their window. Same goes with the polyfill here, as its just querying for a video iframe.

KentoNishi:
That's fax, api shouldn't depend on the player. Holodex won't have the player object either

ChrRubin:
I can pretty easily associate the Watch page player with the YTC via the background messaging API for forwarding keyboard events and the like, but that doesn't solve our issue with the LiveTL API.

Basically for the API code to work on all scenarios it needs to depend on something YTC exposes, rather than the YT player (unless we add a content script to `www.youtube.com/embed/*`, which we could do, but associating a player to a specific chat instance like for MultiView is gonna be tough)

r2dev2:
> Oh boy I'm REALLY not a fan of how hackish this is, and even CodeQL agrees with me on this one lmao.
> 
> Btw the LTL API code you mentioned wouldn't even work in Popout and Embed mode, since there's no player in the context of their window. Same goes with the polyfill here, as its just querying for a video iframe.

We should defnitely migrate the api to something else, now that I think about it, why aren't we using the same timing thing we use for mchad lmao.

But we do need access to the player object for stuff like keyboard shortcuts. Would postMessaging direct events like `volume up` and then have the volume be raised in the embed iframe work better than this?

ChrRubin:
Yeah either using direct event names or even just passing the `KeyboardEvent.code` value to the player script would be 1000% better than invoking dynamic function names. 

Btw unless we wanna get keyboard shortcuts out before smelte, you can probably push/PR it into the smelte branch, since I changed the way the workaround works in there.

r2dev2:
Ok closing this cause it's so bad codeql didn't like it lmao

PR #339, opened 2021-09-11
keyboard shortcuts

### LINKED ISSUE #333: Keyboard shortcuts
Have keyboard shortcuts such as

| Shortcut | Feature        |
|--------------|------------------|
| `up`         | volume up   |
| `down`    | volume down|
| `f`             | full screen |
| `m`           | mute |
| `space/k` | pause |

r2dev2:
Resolves #333

### DIFF
--- a/src/components/KeyboardShortcuts.svelte
+++ b/src/components/KeyboardShortcuts.svelte
@@ -2,6 +2,7 @@
   import { Col, Row, Subheader } from 'svelte-materialify/src';
 
   import { keyboardShortcuts } from '../js/store.js';
+  import TranslatorMode from './settings/TranslatorMode.svelte';
   import KeyboardShortcutInput from './options/KeyboardShortcutInput.svelte';
 
   let timeout = null;
@@ -35,3 +36,5 @@
     </Col>
   </Row>
 {/each}
+
+<TranslatorMode />
--- a/src/components/KeyboardShortcuts.svelte
+++ b/src/components/KeyboardShortcuts.svelte
@@ -1,12 +1,10 @@
 <script>
-  import { Col, Row, Subheader, TextField } from 'svelte-materialify/src';
+  import { Col, Row, Subheader } from 'svelte-materialify/src';
+
   import { keyboardShortcuts } from '../js/store.js';
+  import KeyboardShortcutInput from './options/KeyboardShortcutInput.svelte';
 
   let timeout = null;
-
-  const rules = [
-    key => key.length >= 1 || 'Please input a key'
-  ];
 
   const toDisplayName = actionName => actionName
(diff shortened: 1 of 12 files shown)


src/js/shortcuts.js
```diff
@@ -0,0 +1,34 @@
...
+export const shortcutMap = derived(keyboardShortcuts, $shorts => {
+  const reverseKeyAction = ([action, key]) => [key, action];
+
+  const existingShortcuts = Object.entries($shorts).map(reverseKeyAction);
+  const newShortcuts = Object.entries(defaultShortcuts)
+    .filter(([_key, action]) => $shorts[action] == null)
+    .map(reverseKeyAction);
+
+  return new Map([...existingShortcuts, ...newShortcuts]);
+});
```
ChrRubin: `existingShortcuts` and `newShortcuts` being named the exact reverse of what one would expect is kind of confusing (`existingShortcuts` is the user-set values while `newShortcuts` is the default shortcuts).

This also currently overwrites user-set keys with the default keys, so one can't use a key that was already set for something else in the default keys.

src/components/KeyboardShortcuts.svelte
```diff
@@ -0,0 +1,39 @@
...
+</Subheader>
+
+{#each Object.entries($keyboardShortcuts) as [action, key]}
+  <Row>
+    <Col>
+      <span class="option-label">{toDisplayName(action)}</span>
+    </Col>
+    <Col>
+      <TextField dense on:input={updateKey(action)} {rules} value={key} />
+    </Col>
```
ChrRubin: TextField isn't very user-friendly for changing keyboard shortcuts. The user would need to know the exact `key` value of whatever key they want to use, as well as overall being confusing such as the Spacebar key being an empty ` `.

A better solution would be something that waits for a keydown/keyup event and retrieves the key value for the user.

ChrRubin (changes requested):

src/js/shortcuts.js
r2dev2: Perhaps I should have named `existingShortcuts` as `savedShortcuts` to avoid confusion. Also, `newShortcuts` are just shortcuts that were added in a new update and weren't in the extension storage.

src/components/KeyboardShortcuts.svelte
r2dev2: Done in a67271b

src/js/shortcuts.js
r2dev2: Addressed in cbd9d0b.

KentoNishi (approved):

ChrRubin (approved):

PR #344, opened 2021-09-19
Migrate from getDuration in api live translation

### LINKED ISSUE #338: API code needs to migrate from window.player
The api translation code is reliant upon `window.player` for `getDuration` but `window.player` is no longer accessible.

r2dev2:
This uses the timestamp store instead of `window.player.getDuration()` because the player is not always available.

Fixes #338

### DIFF
--- a/src/js/api.js
+++ b/src/js/api.js
@@ -104,6 +104,6 @@ const apiLiveStream = videoId => derived(
 /** @type {(videoId: String) => Readable<Message>} */
 export const getLiveTranslations = videoId => derived(apiLiveStream(videoId), $data => {
   if ($data?.videoId !== videoId || !enableAPITLs.get()) return;
-  if ($data?.start / 1000 < window.player.getDuration() - 10) return; // if the timestamp of the translation is more than 10 seconds of the timestamp of the player, ignore it
+  if ($data?.start / 1000 < get(timestamp) - 10) return; // ignore if the translation timestamp is > 10s ahead of the player timestamp
   return transformApiTl($data);
 });
--- a/src/js/store.js
+++ b/src/js/store.js
@@ -101,7 +101,7 @@ export const videoSide = derived(videoSideDepends, ([$videoSide, $autoVert, $win
 
 export const updatePopupActive = writable(false);
 export const videoTitle = writable('LiveTL');
-export const timestamp = writable(0);
+export const timestamp = writable(0); // timestamp in seconds
 export const faviconURL = writable('/48x48.png');
 export const availableMchadUsers = writable([]);
 export const spotlightedTranslator = writable(null);


ChrRubin:
The source of the timestamp store value (YT player postMessaging the chat iframe) is not exactly reliable for live streams either. Only chats in the original YT page and in LiveTL mode receive the data, 3rd party sites like Holodex/tools don't forward the postMessage like we do.

And even if they do start forwarding the postMessages, the player only postMessages player progress while the player is playing, and will not send the values while the player is paused etc. For archives its fine since we're not showing TLs while the VOD is paused, but using this value for live streams as well means new TLs from API won't be shown while the player is paused or if the player progress is unavailable.

ChrRubin:
I'm not sure how feasible it is, but it might be better to change the API to use unix epoch for live translations (or maybe it already has unix epoch, idk where to find the API docs). It'd be simpler to use unix epoch when sending TLs to the API too, since with embed mode there's no way 100% reliable way of getting player progress.

r2dev2:
> I'm not sure how feasible it is, but it might be better to change the API to use unix epoch for live translations (or maybe it already has unix epoch, idk where to find the API docs). It'd be simpler to use unix epoch when sending TLs to the API too, since with embed mode there's no way 100% reliable way of getting player progress.

We could have a content script to `postMessage` the timestamp every second from the youtube page. Would that work?

r2dev2:
> I'm not sure how feasible it is, but it might be better to change the API to use unix epoch for live translations (or maybe it already has unix epoch, idk where to find the API docs). It'd be simpler to use unix epoch when sending TLs to the API too, since with embed mode there's no way 100% reliable way of getting player progress.

We could store the initial timestamp and datetime of the livestream and use that to convert from the API's timestamp to unix time.

ChrRubin:
> We could store the initial timestamp and datetime of the livestream and use that to convert from the API's timestamp to unix time.

Or from unix epoch to player time, cause we can't get player info in embed mode reliably for sending TLs to API. Either way it'd probably have to be done API side, don't think the YT player API exposes anything related to stream start time to begin with.

GrumpyBear57:
Have been busy with work things this past week, sorry I haven't been able to comment on this until now. Going to try and address everything all at once. 

--- 

> I'm not sure how feasible it is, but it might be better to change the API to use unix epoch for live translations

No this is not feasible, because part of the primary goal of the API is to store translations for later use. Using a Date/Time format not only doesn't make sense for this purpose, but also would make it difficult to implement for anything outside of a live event. 

> idk where to find the API docs

API docs are at https://docs.livetl.app/api/ (admittedly slightly outdated, but mostly reflective of current status)

> 3rd party sites like Holodex/tools don't forward the postMessage like we do.

This should be solvable by injecting a script to do that forwarding into those sites should it not? 

> [...] using this value for live streams as well means new TLs from API won't be shown while the player is paused or if the player progress is unavailable

When they player is paused, I don't think you'd want to show new translations *anyways*. I can't think of a use case (where popout wouldn't already be the ideal choice) where you'd have a paused stream, but still want new translations to show.

Popout mode can just display everything as it comes in, and sort it so that it's displayed in the appropriate order for the translation timestamps. 

I don't know when the player progress is unavailable outside of 3rd party sites, but as mentioned above, that should be solvable by injecting a script to make it available. 

> Either way it'd probably have to be done API side, don't think the YT player API exposes anything related to stream start time to begin with.

The API is intended to be agnostic against who is sending it data, or when that data is being sent. To me this seems like a responsibility of the client to submit properly formatted data, and is likely much easier for the client to obtain (or already have) the necessary information to properly format the data. 

> We could store the initial timestamp and datetime of the livestream and use that to convert from the API's timestamp to unix time.

I think this is the best solution, but as pointed out, wall time of the stream's start may not be available. One way you could get around this is with the Holofans API (or one of the other Vtuber APIs), which I believe does expose the wall time for the start of the stream. I do recognize this would limit the functionality to only streams tracked by said API, so utilizing the official YT Data API might be the "comprehensive" solution. Checking Vtuber APIs first, then falling back to the official YT API would likely relieve a lot of the ratelimiting concerns that we've spoken about before when considering using it.

ChrRubin:
> This should be solvable by injecting a script to do that forwarding into those sites should it not?

In order to do this we'd need to inject the script into every page that we want to support, since the player iframe itself doesn't expose any of the info within itself. And even then it won't work for pages like Multiview that has multiple player/chat instances, as the postMessage data does not specify exactly which video the data is coming from. `MessageEvent.origin` and `MessageEvent.source` aren't very helpful either.

>> We could store the initial timestamp and datetime of the livestream and use that to convert from the API's timestamp to unix time.
>
> I think this is the best solution, but as pointed out, wall time of the stream's start may not be available.

I agree, seems like the best solution is to get the initial timestamp of the livestream and use it to convert to and from video time. I think mchad is doing this server side since their live translations are in unix epoch but archive translations are in video time. Whether we want to do it client or server side is up for discussion I guess.

GrumpyBear57:
> we'd need to inject the script into every page that we want to support

Do we not already do this for third party sites anyways? I seem to remember having to inject custom scripts for other sites like holotools, to add the buttons for opening/embedding properly. Unless that's changed? 

> I think mchad is doing this server side since their live translations are in unix epoch but archive translations are in video time.

I believe that mchad has a more 'user involved' way to synchronize times, where the they have to manually input/send something. Not sure though, could be wrong (paging @MonMonMnemonics). For ours I'd much rather it be something that "just works", with no setup (other than the one time login and registration) the user has to go through before they start translating.

If absolutely necessary, it could be done serverside, however I still don't feel like it's the appropriate place to do it, due to the aforementioned agnostic compatibility across all types of video. I could add something that might make it easier for the client to make that calculation itself, but all it would be is a proxy/wrapper for another API's endpoints, so don't see much point having the both the API and client do work, when just the client can.

KentoNishi:
@GrumpyBear57 regarding the script injecting part, all of that has changed. we don't inject anything in sites other than youtube, and global injection would be a huge privacy concern for a lot of people so that isn't feasible

GrumpyBear57:
That's fair then, but there *is* a way to request additional permissions "on-demand" (RES does this the first time you try to embed a twitter post into the reddit page) which could be considered. But I honestly don't think it's necessary or worth the effort to work around other sites like holodex/tools anyways (how many translators are translating and watching multiple streams at the same time?). Even if we were to solve the whole timestamp issue, there's going to be other issues that just aren't solvable, as Rubin mentioned already (potentially without realizing the implication):

> Multiview that has multiple player/chat instances, as the postMessage data does not specify exactly which video the data is coming from

That means that you aren't going to be able to send a translation *anyways*, since a translation ***requires*** a Video ID. If you have no way to determine which video the translation is for, being able to determine the proper timestamp isn't going to help you at all. 

It's probably easiest to drop support for LiveTL-extension-based API TL mode with those sites, and (if desired), build it into their site directly, where we would have closer access to the required information. After all, that's part of the reason why the API wrapper exists, so that others can use it. It would mean not supporting literally every site that embeds a YouTube player/chat, but it also means we don't have to dick around with handling every weird scenario that we might come across. That way we can have a more enjoyable translating experience, and would allow more translators to use our API without having to have LiveTL installed.

r2dev2:
Interjecting here, but the pr makes the api translations use the timestamp store which is the same mechanism we use for mchad. Regardless of the way we go about getting the proper timestamp, the api part going to need to use the timestamp store which is what this pr does.

GrumpyBear57:
You're right, this has kind of gotten off topic. Perhaps the conversation can continue in #276, which is the PR for general API integration.

Approving these changes, with the assumption that additional improvements can and will be made at a later time.

GrumpyBear57 (approved):

MonMonMnemonics:
> I believe that mchad has a more 'user involved' way to synchronize times, where the they have to manually input/send something. Not sure though, could be wrong (paging @MonMonMnemonics).

Was surprised I got the notification. Well, mchad actually works both ways, translators can just use it as it is and the timestamp will synchronize automatically. or they can add a manual marker. either way works.

The problem is when translators add notes before the stream actually starts and when the unix epoch timestamp retrieved from the API of the streaming platform is not accurate, that was why I added a button to add a marker manually. Moreover, mchad is a universal tool not only for youtube or streaming platform in particular (It actually started as an on-screen translation tool to be used with zoom or microsoft team or google meet), so there are cases where the streaming platform has no API available to retrieve the timestamp when the stream was started.

PR #349, opened 2021-09-23
Open in LiveTL from embed view button

### LINKED ISSUE #221: No way to open stream in LiveTL after using Embedded TLs

r2dev2:
Adds this button to open in the main LiveTL ui from embed mode
open in ui button

Fixes #221

### DIFF
--- a/src/js/content_scripts/injector.js
+++ b/src/js/content_scripts/injector.js
@@ -1,7 +1,7 @@
 import { mdiOpenInNew, mdiYoutubeTv, mdiIframeArray } from '@mdi/js';
 import { getFrameInfoAsync, createPopup } from '../../submodules/chat/scripts/chat-utils.js';
 import { fixLeaks } from '../../submodules/chat/src/plugins/ytc-fix-memleaks.js';
-import { paramsEmbedDomain, AutoLaunchMode } from '../constants.js';
+import { paramsEmbedDomain, AutoLaunchMode, PostMessage } from '../constants.js';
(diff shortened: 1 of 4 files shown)


KentoNishi (approved):

src/js/content_scripts/injector.js
```diff
@@ -105,6 +105,11 @@ const constructParams = () => {
   return params;
 };
 
+const openLiveTL = () => {
+  window.top.location =
+    chrome.runtime.getURL(`watch.html?${constructParams()}`);
+};
```
ChrRubin, line 111: Couldn't we just export this in `utils.js` and import it whenever its needed instead of doing postMessage?

r2dev2: We'd also need to extract `constructParams()`, but extracting should be fine.

RachitaBas:
it works

ChrRubin:
@RachitaBas Thanks for the feedback. However, we are currently in the middle of a framework migration, so this has been migrated and implemented on the `smelte` branch instead.

@r2dev2 I'm closing this PR in favour of 6e065ee.

PR #352, opened 2021-10-18
Disable spam protection for gigachad users

r2dev2:
- allow non-plebs past the spam protection
- user interface option for enabling taishi bonk
- increase default spam protection trigger to 7 msgs / 10 s

Requested by <@713592330552213554> in the discord

### DIFF
--- a/src/components/SpamProtection.svelte
+++ b/src/components/SpamProtection.svelte
@@ -13,7 +13,10 @@
 
 <div style="margin-top: 1.5rem;">
   <Toggle name="Spam protection" store={enableSpamProtection} />
-  <Toggle name="Authorized user spam protection" store={enableSpecialSpamProtection} />
+  <Toggle
+    name="Authorized user spam protection"
+    store={enableSpecialSpamProtection}
+  />
   {#if $enableSpamProtection}
     <Subheader style="height: 2rem;">
       Hide spammers that send <code>{amount}</code> or more messages within
--- a/src/components/SpamProtection.svelte
+++ b/src/components/SpamProtection.svelte
@@ -1,6 +1,6 @@
 <script>
   import { Subheader } from 'svelte-materialify/src';
-  import { enableSpamProtection } from '../js/store.js';
+  import { enableSpamProtection, enableSpecialSpamProtection } from '../js/store.js';
   import { spammersDetected, spamMsgAmount, spamMsgInterval } from '../js/store.js';
   import Slider from './options/Slider.svelte';
   import Toggle from './options/Toggle.svelte';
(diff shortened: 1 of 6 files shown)


KentoNishi (approved):

KentoNishi:
preliminary approval before a final check by our lord and savior ~lightning mcqueen~ @ChrRubin

src/components/SpamProtection.svelte
```diff
@@ -13,6 +13,10 @@
 
 <div style="margin-top: 1.5rem;">
   <Toggle name="Spam protection" store={enableSpamProtection} />
+  <Toggle
+    name="Authorized user spam protection"
+    store={enableSpecialSpamProtection}
+  />
```
ChrRubin: This can be worded better, its not exactly clear what this setting does. 

Something along the lines of "Include authorized users in spam detection" would be better. The term "authorized users" is still vague, but it would at least get the point across that enabling this setting will enable spam detection even on users that are deemed "authorized".

ChrRubin (changes requested):
Please don't call me the lord and saviour, I think I'm more of a devil than anything lmao

src/components/SpamProtection.svelte
KentoNishi: What about "trusted translators"?

ChrRubin: It includes channel moderators and owner, so its not limited to translators

KentoNishi: This setting only applies to Mchad and LiveTL api no? I don't think moderators and owners matter for that 

Either way, I think "trusted" makes more sense than "authorized"

ChrRubin: https://github.com/LiveTL/LiveTL/blob/57b80ce1ed7ad630cc53505ca9ac8fc6bcf09f80/src/js/constants.js#L69-L70

But yeah "trusted users" would at least be better than "authorized".

r2dev2: Ok so I changed it to "Trusted user spam protection." Is that clearer?

ChrRubin: Not really. "Spam protection" suggests that the extension user is being protected from spam, which works fine as the overall name of the setting/feature.

"Trusted user spam protection" almost sounds like only certain trusted users would benefit from spam protection, which would be very confusing as a concept.

ChrRubin: That or that the spam protection is being supplied by said trusted users, which isn't what's happening here

r2dev2: Should I say "Bonk trusted users" then?

KentoNishi: Make the checkbox say "Bypass spam detection filter for trusted users" checked by default

ChrRubin (approved):

PR #354, opened 2021-10-30
Don't show mchad translations of different language

r2dev2:
## Rationale

Previously, all mchad translations show up regardless of source language. This will be problematic in the new sports festival where there will be -> en translators and -> es translators.

## Tested with

* [kanata stream with -> en translations](https://www.youtube.com/watch?v=ExX_YE6rTGs)
* [holotalk stream with -> es translations](https://www.youtube.com/watch?v=LQcQyLum3Q0)

## Limitations

I have not tested this with live translators yet although if the documentation is correct, it __should__ work. Also, we need to get the translators to all use the correct language tag with their rooms.

### DIFF
--- a/src/__tests__/js/mchad.js
+++ b/src/__tests__/js/mchad.js
@@ -1,0 +1,14 @@
+/* eslint-disable no-undef */
+import { getRoomTagLanguageCode } from '../../js/mchad.js';
+
+it.each([
+  [ 'en', 'en' ],
+  [ 'eng', 'en' ],
+  [ 'Français', 'fr' ],
+  [ '日本語', 'jp' ],
+  [ 'es -> en', 'en' ],
+  [ '', null ],
+  [ 'lalala lalala', null ]
+])('getRoomTagLanguageCode("%s") == %s', (tag, expected) => {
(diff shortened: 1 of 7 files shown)


KentoNishi (approved):
the amount of ts errors we have makes me want to cry, my god

seems like it does the trick but we really gotta do something about the jsdoc fiasco

ChrRubin:
> the amount of ts errors we have makes me want to cry, my god
> 
> seems like it does the trick but we really gotta do something about the jsdoc fiasco

There's something wonderful called `"checkJs": false` lmao. There's really no reason to type check JS before we do a full migrate to TS when we know most of it aren't typed correctly.

Changes seems good nonetheless. Again I'd like to push for getting smelte out to public sooner rather than later so potential bugs reports from public gets fixed BEFORE the sports festival. Can we base this PR on the smelte branch? Aside from the `yarn.lock` changes, the other merge conflicts are pretty minor.

KentoNishi:
i can release this weekend, both hc and ltl

ChrRubin:
Cool, or for simplicity's sake I can just merge either develop or `mchad-language` into smelte later.

ChrRubin (approved):

PR #355, opened 2021-10-31
Manual eslint fixes

ChrRubin:
Gets rid of all the eslint errors. Most of the fixes are manual, so just wanna make sure I didn't break anything.


KentoNishi:
I actually made a svelte typescript template that has eslint configured so it properly lints and autofixes everything in `<script lang="ts">` tags as well as the svelte html markup. i'll probably make a separate plugin sometime so we can have the ideal eslint autofix setup

if u wanna try it, here it is: https://github.com/KentoNishi/svelte-typescript-template

(edit: svelte3/svelte3 plugin doesn't work as nicely with script tags in svelte, my setup actually uses the js/ts lint config for inline scripts)

ChrRubin:
Huh I could've sworn eslint was working fine with the svelte files at some point when I was working on the smelte branch. Weird that it's no longer working now.

ChrRubin:
Actually it seems like eslint itself works on svelte files, but vscode's eslint integration just doesn't run on svelte files now for whatever reason.

KentoNishi:
u gonna try to figure it out or should we just merge?

ChrRubin:
I found what was the problem, had to add this to `settings.json`:
```
"eslint.validate": [
    "javascript",
    "svelte"
]
```
I had this set in project settings instead of global settings before this, which got lost when I moved directories around lmao.

But yeah eslint works fine with svelte files right now from what I can tell, for both checking and autofix.

KentoNishi:
Ayyy epic I approb

It doesn't auto format the markup tho does it? Only the script tags? Just wondering cuz otherwise I spent like 5 hours configuring my template for nothing lmao, that thing auto fixes HTML indentation, order of script HTML css tags, etc. etc.

KentoNishi (approved):

ChrRubin:
Yeah we can probably consider using `@ota-meshi/eslint-plugin-svelte` instead of the official `eslint-plugin-svelte3` if we want eslint to check/fix the svelte markup stuff. I'm more worried about the prettier plugin conflicting with `standard-with-typescript`, which is already pretty opinionated lmao.

PR #357, opened 2021-11-03
Fixed: Messages with only emojis never gets filtered

### LINKED ISSUE #331: Messages with only emojis never gets filtered
Messages that only contain emojis do not contain any text and will never be shown by the LiveTL output, since the text property is required in the messages stores.

This can be undesirable when LiveTL is set to show all mod/owner messages (or even custom user filters), but messages from these users do not get filtered when it only contains emojis.

Eroxl:
Fixes #331

### DIFF
--- a/src/js/sources.js
+++ b/src/js/sources.js
@@ -62,10 +62,10 @@ function attachFilters(translations, mod, ytc) {
     if (!message || isBlacklisted(message)) return;
     const text = message.text.trim();
     const parsed = parseTranslation(text);
-    if (!text) return;
-    if (isTranslation(parsed)) {
+
+    if (text && isTranslation(parsed)) {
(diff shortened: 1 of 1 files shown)


r2dev2 (approved):

r2dev2:
Wait why isn't it queuing the e2e tests?

GrumpyBear57:
> 
> 
> Wait why isn't it queuing the e2e tests?

Probably because of first time contributor settings

r2dev2:
> > Wait why isn't it queuing the e2e tests?
> 
> Probably because of first time contributor settings

I said approve workflow and only 2 checks ran.

r2dev2:
Oh it's because the e2e tests are for push only.

r2dev2:
I ran the e2e tests locally, it seems fine.

PR #366, opened 2021-12-05
fix word break

### LINKED ISSUE #364: [Bug]: Long words don't break properly
### Checklist

- [X] This issue is not a matter of opinion. If it is, I will use the feature suggestion issue template.
- [X] I am still able to reproduce the issue after I reinstall the extension.
- [X] I have clear instructions to reproduce the bug, even on another machine.
- [X] I am not suggesting a feature. If I am, I will use the feature suggestion issue template.

### Describe the bug

Long continuous words doesn't break properly and causes the horizontal scrollbar to show, which messes up the display of all later TLs until a page refresh.

### Steps To Reproduce

1. Open any stream/VOD with TLs in any of the 3 modes.
2. Inspect the TL text and edit it to a single extremely long word without spaces.
3. Watch everything being displayed awfully when the horizontal scrollbar shows up.

(I had an actual example that doesn't include editing with DevTools but Vee edited it to no longer break lmao)

### Expected behavior

Long words should break properly without causing the horizontal scrollbar to show up.

Probably just needs the `break-words` tailwind class somewhere.

### What browsers are you seeing the problem on?

Firefox, Chrome, Microsoft Edge

### Anything else?

This was during the stream but Vee has since edited it:
image

This is with editing TL text via DevTools:
image

KentoNishi:
Fixes https://github.com/LiveTL/LiveTL/issues/364

image

### DIFF
--- a/src/components/MainPane.svelte
+++ b/src/components/MainPane.svelte
@@ -289,7 +289,7 @@
   {#if settingsOpen}
     <slot name="settings" />
   {:else}
-    <div>
+    <div style="max-width: 100%; position: absolute;">
       <MessageDisplay
         direction={$textDirection}
         bind:this={messageDisplay}
--- a/src/components/Message.svelte
+++ b/src/components/Message.svelte
@@ -7,6 +7,7 @@
   import '../css/splash.css';
   import Icon from './common/Icon.svelte';
   import Menu from './common/Menu.svelte';
+  import type { Ltl } from 'ts/typings/types.js';
 
   export let message: Ltl.Message;
   export let hidden = false;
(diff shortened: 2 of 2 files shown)


src/components/Message.svelte
```diff
@@ -7,6 +7,7 @@
   import '../css/splash.css';
   import Icon from './common/Icon.svelte';
   import Menu from './common/Menu.svelte';
+  import type { Ltl } from 'ts/typings/types.js';
```
ChrRubin, line 10: This isn't necessary, and isn't even imported correctly lmao.

The `Ltl` namespace is already included globally thru `tsconfig`.

ChrRubin (changes requested):
Don't worry, Tailwind. I still like you.

src/components/MainPane.svelte
```diff
@@ -289,7 +289,7 @@
   {#if settingsOpen}
     <slot name="settings" />
   {:else}
-    <div>
+    <div style="max-width: 100%; position: absolute;">
```
ChrRubin, line 292: Absolute just makes it look like this, even on bottom text direction:

image

r2dev2: This seems to be the test that failed. The tests actually caught something. I'm so proud.

PR #377, opened 2021-12-22
Ability To Select Multiple Filter Languages

### LINKED ISSUE #361: [Feature]: Select multiple filter languages
### Checklist

- [X] I am not reporting a bug. If I am, I will use the bug report issue template.

### Is your feature request related to a problem? Please describe.

While it is technically possible to show TLs from multiple languages from YTC with custom filters, Mchad TLs shown are based on the filter language selected, making it impossible for users to view Mchad TLs from different languages at the same time.

### Describe the solution you'd like

Checkbox selection for the filter languages, and show TLs for all the languages that are selected.

### Describe alternatives you've considered

_No response_

### Anything else?

_No response_

Eroxl:
# About
Added the ability to select multiple filter languages for translations from a list of check boxes of all the valid translation tags.

# Example
Screen Shot 2021-12-22 at 5 58 47 PM

Closes #361


Eroxl:
Sorry this isn't actually ready yet so don't merge it.

ChrRubin:
Is this ready for review? Just wanna double check cause there was a comment before about it not being ready

Eroxl:
@ChrRubin Should be ready now sorry about that.

r2dev2:
@ChrRubin @KentoNishi @Eroxl added the request for review so we remember to review it.

src/components/settings/TranslatorMode.svelte
```diff
@@ -38,7 +43,8 @@
     {#if $doAutoPrefix}
       <TextField
         bind:value={$autoPrefixTag}
-        label="Tag to prepend ($filterLang is replaced by your filter language)"
+        label="Tag to prepend ($filterLang is replaced by your first filter language)"
+        rules={autoPrefixRules}
       />
```
ChrRubin: Should have a dropdown to select a specific language tag instead. This currently makes it unclear which language exactly is being used, and a translator might for example want to still see EN translations but send translations in another language.

ChrRubin (changes requested):

src/js/store.js
```diff
@@ -52,7 +52,7 @@ export const defaultShortcuts = {
 };
 
 // Settings
-export const language = SS('language', 'English');
+export const languages = SS('languages', ['English']);
```
r2dev2: This will cause migration errors as some people have already selected a language other than English and the choice will not be reflected in the updated app. Can you make it so that the `language` store contains the primary language (the one to be used in the translator mode auto-prefix) and make the `languages` store update its value based on the initial `language` store value? I was thinking something along the lines of

```js
await language.loaded;
await hasDoneLanguageMigration.loaded;
if (!hasDoneLanguageMigration) {
  languages.set([language.get()]);
}
```

src/js/translator-mode.js
```diff
@@ -286,4 +286,4 @@ function getCaretCharOffset(element) {
   return caretOffset;
 }
 
-const langTag = () => autoPrefixTag.get().replace(/\$filterLang/gi, languageNameCode[language.get()].code);
+const langTag = () => languages.length <= 0 ? autoPrefixTag.get() : autoPrefixTag.get().replace(/\$filterLang/gi, languageNameCode[languages.get()[0]].code);
```
r2dev2: Instead of `languageNameCode[languages.get()[0]]`, use the repurposed `language` store.

src/js/filter.js
```diff
@@ -157,6 +157,15 @@ export function isLangMatch(textLang, currentLang) {
   ));
 }
 
+/**
+ * @param {String} textLang
+ * @param {{ code: String, name: String, lang: String }[]} currentLang
+ * @returns {Boolean}
+ */
+export function isLangListMatch(textLang, currentLangs) {
+  return currentLangs.map((currentLang) => isLangMatch(textLang, currentLang)).includes(true);
```
r2dev2: use `.some()` instead of `.map().includes(true)`

r2dev2 (changes requested):

Eroxl:
Busted up my computer so I'm gonna have to finish this in a bit sorry.

src/js/store.js
Eroxl: @r2dev2 Ok can you check out the changes I made I just used the empty languages store instead of a separate store to detect migration. Not sure if there's an issue with that or not so lmk what you think.

src/js/store.js
```diff
@@ -137,6 +138,18 @@ export const presetStores = [
...
 ];
 
+// -=- Language Migration -=-
+(async () => {
+  const language = SS('language', 'English');
+
+  await language.loaded;
+  await languages.loaded;
+
+  if (!languages.get().length <= 0) {
```
r2dev2: This will only run when `languages` is not empty which is the opposite of what you wanted. Also, I think that a separate store which stores whether the migration happened would work better as the user may want to get no translations from YouTube chat.

src/js/translator-mode.js
```diff
@@ -286,4 +286,4 @@ function getCaretCharOffset(element) {
   return caretOffset;
 }
 
-const langTag = () => autoPrefixTag.get().replace(/\$filterLang/gi, languageNameCode[language.get()].code);
+const langTag = () => autoPrefixTag.get().replace(/\$filterLang/gi, languageNameCode[prefixTagReplacementLanguage].code);
```
r2dev2: `prefixTagReplacementLanguage` is a `SyncStore`, you need to call `.get()` on it

src/components/settings/TranslatorMode.svelte
```diff
@@ -29,6 +30,11 @@
   }
 
   onMount(() => setTimeout(cleanUpMacros));
+
+  const autoPrefixRules = [{
+    error: 'You have no selected translation languages',
+    assert: (value) => !(/\$filterLang/gi.test(value)),
+  }];
```
r2dev2: Rules are not needed in these settings anymore since `$filterLang`  always has a replacement.

r2dev2 (changes requested):
User translation language migrations don't work.

src/js/store.js
r2dev2: I think that a separate store which stores whether the migration happened would work better as the user may want to get no translations from YouTube chat.

src/components/settings/TranslatorMode.svelte
```diff
@@ -38,8 +39,14 @@
     {#if $doAutoPrefix}
       <TextField
         bind:value={$autoPrefixTag}
-        label="Tag to prepend ($filterLang is replaced by your filter language)"
+        label="Tag to prepend ($filterLang is replaced by the language selected below)"
       />
+      <Dropdown
+        store={prefixTagReplacementLanguage}
+        items={languagesInfo.map((language) => language.name)}
```
r2dev2: Instead of `=> language.name`, I think `=> language.lang` is intended. I think that *should* fix the auto-prefix error.

r2dev2 (changes requested):
It keeps migrating from the previous settings even when you change them (eg. if I previously had my language set to japanese, then download the new version and uncheck japanese and then check english, it will go back to just japanese after a relaunch.

Also, for some reason, the translator mode auto-prefix doesn't work with languages other than English?

Eroxl:
@r2dev2 Sorry about that migration issue i just forgot to change the boolean once you migrated.

r2dev2 (approved):

KentoNishi (approved):

src/components/settings/TranslatorMode.svelte
```diff
@@ -38,8 +39,14 @@
...
         bind:value={$autoPrefixTag}
-        label="Tag to prepend ($filterLang is replaced by your filter language)"
+        label="Tag to prepend ($filterLang is replaced by the language selected below)"
       />
+      <Dropdown
+        store={prefixTagReplacementLanguage}
+        items={languagesInfo.map((language) => language.lang)}
+      >
+        Auto Prefix Tag
+      </Dropdown>
```
ChrRubin: Should add a `name` to the dropdown, right now it looks awkwardly empty:

image

src/components/settings/TranslatorMode.svelte
```diff
@@ -1,13 +1,14 @@
 <script>
   import { onMount } from 'svelte';
   import CustomMacro from '../options/CustomMacro.svelte';
-  import { macros, doAutoPrefix, doTranslatorMode, autoPrefixTag, macroTrigger } from '../../js/store.js';
-  import { isAndroid } from '../../js/constants.js';
+  import { macros, doAutoPrefix, doTranslatorMode, autoPrefixTag, macroTrigger, languages, prefixTagReplacementLanguage } from '../../js/store.js';
```
ChrRubin: `languages` is unused.

src/components/settings/CommonFilterCards.svelte
```diff
@@ -2,33 +2,33 @@
...
   import { languageNameValues } from '../../js/constants.js';
   import BlockedUsers from './BlockedUsers.svelte';
-  import Checkbox from '../common/CheckboxStore.svelte';
-  import Dropdown from '../common/DropdownStore.svelte';
+  import Checkbox from '../common/Checkbox.svelte';
+  import CheckboxStore from '../common/CheckboxStore.svelte';
   import Card from '../common/Card.svelte';
   export let div: HTMLElement;
+
+  let validLanguages: {[key: string]: boolean} = Object.fromEntries($languages.map((language) => [language, true]));
```
ChrRubin: `validLanguages` is never reassigned. Should use `const`.

ChrRubin (changes requested):
Just a small UI suggestion and some eslint stuff.

src/components/settings/TranslatorMode.svelte
Eroxl: Not 100% sure on how to phrase it so check it out @ChrRubin

ChrRubin: Something like "Prefix language" should be straightforward enough I guess

ChrRubin (approved):

PR #388, opened 2022-01-23
update hyperchat selectors

r2dev2:
Tests are broken, update hyperchat button selectors. I will merge when ci shows that tests pass.


r2dev2:
Poggers, tests work again.

PR #407, opened 2022-04-07
fix: check for ltl visibility before scrolling msg into view

r2dev2:
Fixes #403

PR #408, opened 2022-04-08
Fix: don't scroll parent window when scrolling messages

### LINKED ISSUE #403: [Bug]: Parent page scrolls to filtered message on Firefox
### Checklist

- [X] This issue is not a matter of opinion. If it is, I will use the feature suggestion issue template.
- [X] I am still able to reproduce the issue after I reinstall the extension.
- [X] I have clear instructions to reproduce the bug, even on another machine.
- [X] I am not suggesting a feature. If I am, I will use the feature suggestion issue template.

### Describe the bug

When using Embed Mode on Firefox, `scrollIntoView` will also scroll the entire parent page to the specified element. This is especially apparent when using YouTube's theater mode, where the chat and LiveTL are off-screen.

This might be a problem with Firefox's implementation of `scrollIntoView`, based on [this currently unanswered SO post](https://stackoverflow.com/questions/70816641/scrollintoview-ancestor-frame-scrolling-in-firefox-but-not-chrome).

### Steps To Reproduce

1. Open a stream/VOD with YT theater mode.
2. Launch LiveTL in embed mode.
3. Wait for a filtered message.

### Expected behavior

The parent page doesn't scroll to the filtered message.

### What browsers are you seeing the problem on?

Firefox

### Anything else?

https://gfycat.com/ko/boldunsightlyherald

r2dev2:
Fixes #403 
Closes #407 

This one uses `.scrollTo` instead of intersection observer shenanigans.

### DIFF
--- a/src/components/MainPane.svelte
+++ b/src/components/MainPane.svelte
@@ -334,7 +334,7 @@
           ? 'arrow_upward'
           : 'arrow_downward'}
         on:click={() => {
-          messageDisplay.scrollToRecent();
+          messageDisplay.scrollToRecent(true);
           updateWrapper();
         }}
         filled
--- a/src/components/MainPane.svelte
+++ b/src/components/MainPane.svelte
@@ -334,7 +334,7 @@
           ? 'arrow_upward'
           : 'arrow_downward'}
         on:click={() => {
-          messageDisplay.scrollToRecent(true);
+          messageDisplay.scrollToRecent();
           updateWrapper();
         }}
         filled
--- a/src/js/store.js
+++ b/src/js/store.js
@@ -209,3 +209,4 @@ export const anyRecordingShortcut = writable(false);
 export const videoShortcutAction = writable('');
 export const currentlyEditingPreset = writable(-1);
 export const promptToShow = writable(/** @type {String[]} */ ([]));
+export const scrollTo = writable({ top: 0 });
--- a/src/js/store.js
+++ b/src/js/store.js
@@ -209,4 +209,4 @@ export const anyRecordingShortcut = writable(false);
 export const videoShortcutAction = writable('');
 export const currentlyEditingPreset = writable(-1);
 export const promptToShow = writable(/** @type {String[]} */ ([]));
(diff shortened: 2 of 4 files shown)


src/js/store.js
```diff
@@ -209,3 +209,4 @@ export const anyRecordingShortcut = writable(false);
 export const videoShortcutAction = writable('');
 export const currentlyEditingPreset = writable(-1);
 export const promptToShow = writable(/** @type {String[]} */ ([]));
+export const scrollTo = writable({ top: 0 });
```
ChrRubin: This should be typed with `ScrollToOptions`.

src/components/MainPane.svelte
```diff
@@ -334,7 +334,7 @@
           ? 'arrow_upward'
           : 'arrow_downward'}
         on:click={() => {
-          messageDisplay.scrollToRecent();
+          messageDisplay.scrollToRecent(true);
```
ChrRubin: Smooth scrolling has the same problem as when I tried to implement it a long time ago, it won't scroll all the way if new messages show up while the scroll animation is running.

Granted, I don't think this will happen often during proper usage of LTL, but it was decided previously to just scrap the idea of smooth scrolling to prevent this.

https://user-images.githubusercontent.com/20059164/162567382-4098c84e-51f6-4208-806e-672ca42f494d.mp4

ChrRubin (changes requested):

src/js/store.js
r2dev2: Fixed in [5ccc1ca](https://github.com/LiveTL/LiveTL/pull/408/commits/5ccc1ca8919c134aa56ff174f67e3bcef3e167b4)

src/components/MainPane.svelte
r2dev2: Fixed in [5ccc1ca](https://github.com/LiveTL/LiveTL/pull/408/commits/5ccc1ca8919c134aa56ff174f67e3bcef3e167b4)

ChrRubin (approved):

KentoNishi (approved):

PR #409, opened 2022-04-14
Remove message duplication & Read aloud duplication

r2dev2:
For some reason, the same message with the same `messageId` is being sent twice. This uses the `messageId` to check for duplicate messages. In the console, HyperChat was complaining about duplicates in the keyed each block which makes me believe we need to do something similar in HyperChat.

Bug was reported by gregsusername#6281 in ltlcord in #tech-support

### DIFF
--- a/src/js/sources-aggregate.js
+++ b/src/js/sources-aggregate.js
@@ -157,6 +157,6 @@ export const captionText = readable(defaultCaption, set => {
 // hmr
 if (window.unsubDictation) window.unsubDictation();
 window.unsubDictation = captionText.subscribe($txt => {
-  console.debug(`[LiveTL] Caption text:`, $txt);
-  $txt.text !== defaultCaption.text && checkAndSpeak($txt)
+  console.debug('[LiveTL] Caption text:', $txt);
+  $txt.text !== defaultCaption.text && checkAndSpeak($txt);
 });
--- a/src/js/sources-aggregate.js
+++ b/src/js/sources-aggregate.js
@@ -145,7 +145,10 @@ export const displayedMessages = derived(dispDepends, dispTransform);
 export const captionText = readable(defaultCaption, set => {
   let text = defaultCaption;
   return displayedMessages.subscribe($msgs => {
-    if ($msgs[$msgs.length - 1]?.text != null && $msgs[$msgs.length - 1]?.text !== text) {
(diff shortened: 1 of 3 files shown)


src/js/sources-aggregate.js
```diff
@@ -145,12 +145,18 @@ export const displayedMessages = derived(dispDepends, dispTransform);
 export const captionText = readable(defaultCaption, set => {
   let text = defaultCaption;
   return displayedMessages.subscribe($msgs => {
-    if ($msgs[$msgs.length - 1]?.text != null && $msgs[$msgs.length - 1]?.text !== text) {
+    if ($msgs[$msgs.length - 1]?.text != null &&
+      $msgs[$msgs.length - 1]?.text !== text &&
```
KentoNishi, line 149: `$msgs[$msgs.length - 1]`
what if there's a message between the duplicated message 1 and 2? i feel like this is a patch that isn't fixing the fundamental problem

KentoNishi (changes requested):

src/js/sources-aggregate.js
KentoNishi: image

KentoNishi:
@r2dev2 i fixed the self-send message duplication issue, can you strip the message duplication removal and just keep the read-aloud duplication removal part?

src/js/sources-aggregate.js
r2dev2: This is the read aloud duplication, it is only supposed to check against the last message.

r2dev2:
> @r2dev2 i fixed the self-send message duplication issue, can you strip the message duplication removal and just keep the read-aloud duplication removal part?

The current pr adds `messageId` to the check for duplicate messages and adds `messageId` to the check for duplicate read aloud. I think it's a good idea to also check the `messageId` prop for duplicate messages as it doesn't hurt and the other check is needed to make sure read aloud doesn't duplicate.

KentoNishi:
i'll await @ChrRubin's evaluation

ChrRubin (approved):
I can't seem to reproduce read aloud duplicating messages other than those caused by the fake instant message sending bug (that should be now fixed on HC side), but the changes here don't hurt.

KentoNishi (approved):

PR #412, opened 2022-05-07
mchad_to_tldex migration

MonMonMnemonics:
fully working migration to tldex
- 1 new constant in constant.js to add holodexapi adderss
- 1 change in message.svelte from mchad TL -> tldex check mark
- not really small changes in mchad.js

the migration is scheduled to be by the end of golden week, so most probably on next monday

### DIFF
--- a/src/components/Message.svelte
+++ b/src/components/Message.svelte
@@ -103,7 +103,7 @@
         <span class="{nameColorClass} inline-block align-middle">{message.author}</span>
         {#if mchad}
           <span class="bg-gray-700 px-1 rounded inline-block text-gray-300 align-middle">
-            <Icon block={false} class="inline align-middle" small>check_circle</Icon> Mchad TL
+            <Icon block={false} class="inline align-middle" small>check_circle</Icon> TLdex
           </span>
         {/if}
         <span class="inline-block align-middle">{timestamp}</span>
--- a/src/js/constants.js
+++ b/src/js/constants.js
@@ -132,6 +132,7 @@ languagesInfo.forEach(i => (languageConversionTable[createLangSelectionName(i)] 
 languagesInfo.forEach(lang => (languageNameCode[lang.lang] = lang));
 
 export const MCHAD = 'https://repo.mchatx.org';
+export const Holodex = 'https://holodex.net/api/v2';
 
 const params = new URLSearchParams(window.location.search);
 export const paramsYtVideoId = params.get('ytVideo');
--- a/src/js/mchad.js
+++ b/src/js/mchad.js
@@ -1,109 +1,64 @@
-import { MCHAD, AuthorType, languagesInfo } from './constants.js';
+import { MCHAD, Holodex, AuthorType, languagesInfo, languageNameCode } from './constants.js';
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 import * as Ty from './types.js';
 import { derived, readable } from 'svelte/store';
-import { enableMchadTLs, mchadUsers } from './store.js';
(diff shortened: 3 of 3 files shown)


ChrRubin:
Since the whole script is pretty much rewritten, we might wanna just do it in TypeScript instead.

Also, would this migration affect older archived translations?

MonMonMnemonics:
everything under src/js where mchad.js is located is all in js anyway. You can do that later if you want but for now just keep it as it is,

mchad has been sending translation to holodex as well for a long time now, so past archives should be ok

ChrRubin:
Are super old archives gonna be sent to TLdex too? For example https://youtu.be/tTyLbWrGKNs which was probably before Holodex was even a thing.

MChad archives have translations to this VOD but not on TLdex.

MonMonMnemonics:
June 2020, I feel old now.... it has been that long huh.

There is an "import from mchad" button in https://staging.holodex.net/scriptmanager. Unfortunately translators have to do it manually since there's no way to connect mchad and holodex account directly. But hey, it's just 3 clicks process and I'll at least urge them to do so.

KentoNishi:
Can we have both mchad and tldex connected up in that case? It'd be bad if previously accessible tls became inaccessible becuz of this vendor switch. Just to make sure, this effectively means mchad is no more, and tldex is taking over, right?

MonMonMnemonics:
You can use old code to get tl from mchad and the new code to get tl from tldex, then work out the duplication. 

But just to let you know, mchad is effectively closing soon. Ph.D in particle physics is hardcore, last month alone I was only able to work on the migration in two occasions and one of them was thanks to golden week.

KentoNishi:
damn good luck with your defense, whenever it is

ChrRubin:
> But just to let you know, mchad is effectively closing soon. Ph.D in particle physics is hardcore, last month alone I was only able to work on the migration in two occasions and one of them was thanks to golden week.

Will it still be up for translators to use or are they expected to move to TLdex as well? Also will the old archives that are not on TLdex still be accessible?

> You can use old code to get tl from mchad and the new code to get tl from tldex, then work out the duplication.

Assuming the old archives will still be available, we can just add TLdex as a seperate source and deduplication *should* work as is. Although I wonder if there'll be weirdness with having TLs with the exact same timestamps from 2 sources, such as one TL being marked as from MChad and the next one being marked as from TLdex.

MonMonMnemonics:
I'm planning to keep it up for 1-2 month during the transfer to give translator time to import their scripts as well as to clear up bugs. After that, it will be completely shut down except for the bot since some people asked me to keep that up.

ChrRubin:
I've made a new branch on  [`tldex`](https://github.com/LiveTL/LiveTL/tree/tldex) ([diff](https://github.com/LiveTL/LiveTL/compare/tldex)) and copied the modified script to a separate file so TLdex can be its own source alongside MChad. This will let us still show old translations that are on MChad but not on TLdex, as long as the MChad archives stay up.

If that's the path we wanna take, I can merge it into this branch or open a new PR. If not, the changes here seems fine to me.

MonMonMnemonics:
I'll leave the decision to the development team.

Just a note that currently holodex only have websocket and through my discussion with them, liveTL is proxying through mchad to reduce the load and I'm planning to downsize the server to bare minimum.

KentoNishi:
heads up, mon figured out automated migration with the holodex devs, so it's safe to transition all mchad functionality to dex now! @ChrRubin do u have time to whip up a ts version of this? im getting crushed with covid, finals, and graduation stuff...

KentoNishi:
https://github.com/LiveTL/LiveTL/pull/413

PR #413, opened 2022-05-23
tldex

KentoNishi:


KentoNishi:
mchad is now down and disconnected, so there isn't much of a reason to keep the code anymore, but oh well

r2dev2 (approved):
Gonna just set the `enableTldexTl` to be equal to the `enableMCHADTl` store so that we keep the user preference of those who disabled mchad.

ChrRubin (approved):
One side effect to note is that YTC translations might also get the TLdex badge depending on which message is received first, since TLdex catches YTC TLs too.

r2dev2:
> One side effect to note is that YTC translations might also get the TLdex badge depending on which message is received first, since TLdex catches YTC TLs too.

Should we play with the verified parameter in the tldex url then? What does that verified number do?
verified

ChrRubin:
> Should we play with the verified parameter in the tldex url then? What does that verified number do? verified

I would guess that's for the YT verified badge, but chat messages is not on their public API reference.

KentoNishi:
good enough for now, release will be tomorrow

PR #415, opened 2022-06-14
Mark tl as spam with user confirmation

r2dev2:
Basically does:
```
@r2dev2 just a quick note on what I mean

make it so that when the user first opens this update, the block list is cleared (so probably do it in the changelog component, but keep in mind that can be opened on demand)

add a confirmation dialog when a user is detected for spam, only add to the block list if the user confirms
```
-- @KentoNishi 

- [x] migration to clear spam block list
- [x] logic for user-confirmed spam blocking
- [x] ui for user-confirmed spam blocking

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -32,7 +32,9 @@ module.exports = {
         ],
         '@typescript-eslint/prefer-nullish-coalescing': 'off',
         // Causes false positives with reactive and auto subscriptions
-        '@typescript-eslint/strict-boolean-expressions': 'off'
+        '@typescript-eslint/strict-boolean-expressions': 'off',
+        'no-unused-expressions': 'off',
+        'no-sequences': 'off',
       }
     }
   ],
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "LiveTL",
-  "version": "8.2.11",
+  "version": "8.3.0",
   "description": "Get live translations for YouTube and Twitch streams, crowdsourced from multilingual viewers!",
   "scripts": {
     "build": "webpack --mode development",
--- a/src/components/MainPane.svelte
+++ b/src/components/MainPane.svelte
@@ -29,6 +29,7 @@
   import UpdateComponent from './Updates.svelte';
   import MessageDisplay from './MessageDisplay.svelte';
   import FeaturePrompt from './FeaturePrompt.svelte';
+  import SpammerPrompt from './SpammerPrompt.svelte';
   import { displayedMessages } from '../js/sources-aggregate.js';
   import dark from 'smelte/src/dark';
   import Button from './common/IconButton.svelte';
@@ -197,6 +198,7 @@
   renderWidth={$screenshotRenderWidth}
 />
 <UpdateComponent />
+<SpammerPrompt />
 
 <div
   class="flex flex-row gap-2 absolute right-0 p-1 z-20 flex-wrap"
--- a/src/components/SpammerPrompt.svelte
+++ b/src/components/SpammerPrompt.svelte
@@ -74,7 +74,9 @@
     User
     <a
       class="text-primary-400 hover:underline"
-      href="https://www.youtube.com/channel/{spammerInQuestion.authorId}">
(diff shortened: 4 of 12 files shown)


r2dev2:
@KentoNishi can u do the ui for the spam blocking? I'm assuming it's a `<Dialog />`, but unsure and it's 1 am lmao. Modify `src/components/SpammerPrompt.svelte` for the ui. `spammerInQuestion` is either `null` or an author object. `markAsSpammer` is the callback for marking the user as a spammer and `markAsInnocent` will whitelist them from the spam detection.

KentoNishi:
it's bwoken

image

KentoNishi:
@r2dev2 i fixed reactivity, but im so confused

why do we have both `channelFilters` and `spammersDetected`??? why are the blocklists separate? can u change it so the spam detection thing uses channelFilters instead, merging all blocking functionality into one?

r2dev2:
> @r2dev2 i fixed reactivity, but im so confused
> 
> why do we have both `channelFilters` and `spammersDetected`??? why are the blocklists separate? can u change it so the spam detection thing uses channelFilters instead, merging all blocking functionality into one?

Ok, I merged the spam functionality to use the normal `channelFilters` and `mchadUsers` stores. It also now uses a separate `spammersWhitelisted` store (integrated nicely cause I was deriving such a store later on lmao).

KentoNishi:
@r2dev2 im confused again. how do i differentiate between tldex and chat users in `spammersWhitelisted`? can u make it so `spammersWhitelisted` stores the source? why does `channelFilters` have `{ name: '', blacklist: false, whitelist: false }`, even tho `whitelist` is never used? can we just have one place where all user whitelists and blacklists are stored as an array of `{ userId: string; source: 'chat' | 'tldex'; type: 'blacklist' | 'whitelist' }` or something like that? i really hate that we have multiple places where user stuff is stored, we NEED to merge everything into one

ChrRubin:
I suggest adding a "default behaviour" option if people don't want to respond to the dialog every time, something like a "remember my choice" checkbox.

Defaulting to "block" will just be our previous behaviour, and defaulting "ignore" should just disable spam protection. The "whitelist" button should be greyed out when selecting a default since it doesn't really make sense.

Also would need to add a toggle to the spam protection settings for "always block" if this is added. "Always ignore" will just disable spam protection so we don't need an extra setting for that.

r2dev2:
> @r2dev2 im confused again. how do i differentiate between tldex and chat users in `spammersWhitelisted`? can u make it so `spammersWhitelisted` stores the source? why does `channelFilters` have `{ name: '', blacklist: false, whitelist: false }`, even tho `whitelist` is never used? can we just have one place where all user whitelists and blacklists are stored as an array of `{ userId: string; source: 'chat' | 'tldex'; type: 'blacklist' | 'whitelist' }` or something like that? i really hate that we have multiple places where user stuff is stored, we NEED to merge everything into one

Spammer whitelist is different from an overall whitelist, channelFilter whitelist is supposed to allow all messages from the channel. Although looking at the code, idk if we're even implementing that part. But yea, we should refactor all the user information into one store. That may take a while so I'll do it in a separate pr (perhaps to this branch).

Should we implement @ChrRubin 's suggestion of default behavior and then release for now? Or should we just make a small temporary edit to whitelist taishi from spam detection and then do the whole user whitelist refactor?

KentoNishi:
I have to add a whitelist dropdown, but for now I think we should probably just release. Rubin's suggestion and the refractor can come later ig

ChrRubin:
We should at least have a way to edit the whitelist before releasing, right now there's no way to remove whitelists.

And yeah the whitelist boolean in `channelFilters` is unused, I don't think it'll cause problems if we just remove it. JS only complains if something doesn't exist, not when there's extra values after all.

KentoNishi:
image

KentoNishi (approved):

KentoNishi:
@ChrRubin when u wake up pls just quickly review, i wanna get this update out today thx

ChrRubin (approved):

PR #417, opened 2022-06-19
Handle YTC TLs from TLdex

ChrRubin:
Overall this will prevent the TLdex badge from showing on YTC TLs that are collected by TLdex, specifically anything that has a `channel_id` value. This also assumes TLdex TLers as trusted users in terms of anti spam. 

On archives, "YTC TLs" from TLdex are treated as normal YTC TLs. This is mainly because the MChad archive migration process had set their internal UUID as `channel_id` in TLdex, which means if we just ignore anything with `channel_id` those will stop showing up in LiveTL (see e2e errors on first 2 commits). Showing old MChad TLs as normal YT TLs is at least better than not showing them at all. It's also just beneficial against YT purging archived chat messages.

On live streams, YTC TLs from TLdex are just ignored. I doubt the TLdex scrapper will catch that much more compared to users' browser chats in real time, and there's not really an easy way to prevent this from happening aside from asking MonMon to pass through all of the user type booleans from TLdex as well:

image

@r2dev2 gonna need some help fixing the e2e script. Pretty sure its been complaining ever since we moved from Mchad to TLdex lmao.

### DIFF
--- a/e2e/test_ext.py
+++ b/e2e/test_ext.py
@@ -37,6 +37,7 @@ configure(
 chilled_cow = "https://www.youtube.com/watch?v=5qap5aO4i9A"
 peko_kiara = "https://www.youtube.com/watch?v=c747jYku6Eo"
 mio_phas = "https://www.youtube.com/watch?v=h7F4XJh41uo"
+mio_2 = "https://www.youtube.com/watch?v=4I2iZahIDNg"
 phas = "Phasmophobia"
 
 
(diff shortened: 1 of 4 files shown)


r2dev2:
@ChrRubin

So, should I remove the checks for tldex badge in e2e for now? Also, maybe we can identify mchad users. Mon manually approved all mchad users so there must be a list with all the ids of mchad users. Maybe we can ask him for the list?

ChrRubin:
On the old Phasmophobia VOD for example no longer shows any badge, so either we stop checking for the badge on it or find a newer VOD with TLdex TLs for e2e. You can see the ones failing due to no badge on the test-failures discord channel.

Regarding the mchad list, this only affects archives that used to be only on mchad and was then migrated to tldex when mchad is closing (really really old archives). Anything since TLdex added mchad support on their end should not be affected. I don't think it's too much of a big deal.

KentoNishi:
@r2dev2 change the e2e tests and ill merge

r2dev2:
> @r2dev2 change the e2e tests and ill merge

They work nyow

r2dev2 (approved):

PR #426, opened 2022-08-13
Fix placement of message menu depending on location in screen

r2dev2:
Taishi reported bug, I forgor to make an issue so I'll just put it here:

In some layout configurations (as shown below), the message menu is hidden by other panels

message menu hidden

This pr fixes this by checking the client rectangles against an optionally given container element instead of just the document body.

### DIFF
--- a/src/components/common/Menu.svelte
+++ b/src/components/common/Menu.svelte
@@ -56,23 +56,27 @@
       return;
     }
 
+    const containerEl = document.querySelector(containerQuery) ?? document.body;
     const containerRect = containerEl.getBoundingClientRect();
     const activatorRect = activator.getBoundingClientRect();
     const offsetY = activator.clientHeight + 5;
-    if (activatorRect.top - listDiv.clientHeight > containerRect.top) {
+    if (activatorRect.top - listDiv.clientHeight > Math.max(containerRect.top, 0)) {
+      // put on the top of activator
       offsetYStyle = `bottom: ${offsetY}px;`;
     } else {
+      // put on the bottom of activator
       offsetYStyle = `top: ${offsetY}px;`;
     }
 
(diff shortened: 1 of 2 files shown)


src/components/common/Menu.svelte
```diff
@@ -54,21 +56,23 @@
...
 
-    if (activatorRect.right + listDiv.clientWidth > windowInnerWidth) {
+    if (activatorRect.right + listDiv.clientWidth > containerRect.right) {
       offsetX = 'right-0';
     } else {
       offsetX = 'left-0';
     }
   };
 
+  $: containerEl = document.querySelector(containerQuery) ?? document.body;
```
ChrRubin: Might want to move this to right before `containerEl.getBoundingClientRect()`. The element is yeeted whenever the settings page is opened/closed, so menus on old messages will break while new ones are fine:

https://user-images.githubusercontent.com/20059164/184477711-7a3cac6e-2ce2-4182-822f-1323ef39888c.mp4

ChrRubin (changes requested):

src/components/common/Menu.svelte
r2dev2: Fixed in [deb6df7](https://github.com/LiveTL/LiveTL/pull/426/commits/deb6df70e4ea9b9eb5d22311ccaea5644462ccef)

r2dev2:
Btw, I found another issue in which if the message is at the top of the MessageDisplayWrapper and there are a lot of messages, clientRect positions can have negative values which throws off the calculations. I fixed it in [deb6df7](https://github.com/LiveTL/LiveTL/pull/426/commits/deb6df70e4ea9b9eb5d22311ccaea5644462ccef) as well

menu above screen

ChrRubin (approved):

KentoNishi (approved):

PR #450, opened 2023-09-07
Fix Translator Mode

r2dev2:
YT added yet another element with id of input (a g#input) so adjust for that.

Also, explicitly check for null errors to avoid crashing.


KentoNishi (approved):
seems good

PR #453, opened 2023-11-01
feat: Enable Holodex, misc: Options Page

XangelMusic:
Created a setting to enable LiveTL on Holodex in Advanced > Additional sites in Options.

Set open_in_tab to false in manifest for Options Page (Firefox) and Embedded Options (Chrome).

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -2,7 +2,7 @@
   "name": "LiveTL - Translation Filter for Streams",
   "options_ui": {
     "page": "options.html",
-    "open_in_tab": true
+    "open_in_tab": false
   },
   "permissions": [
     "storage",
--- a/src/ts/content_scripts/injector.ts
+++ b/src/ts/content_scripts/injector.ts
@@ -1,7 +1,7 @@
 import { mdiOpenInNew, mdiYoutubeTv, mdiIframeArray } from '@mdi/js';
 import { getFrameInfoAsync, createPopup, isValidFrameInfo } from '../../submodules/chat/src/ts/chat-utils';
 import { paramsEmbedDomain, AutoLaunchMode } from '../../js/constants.js';
-import { autoLaunchMode } from '../../js/store.js';
+import { autoLaunchMode, holodexEnabled } from '../../js/store.js';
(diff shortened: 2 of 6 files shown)


src/manifest.json
```diff
@@ -2,7 +2,7 @@
   "name": "LiveTL - Translation Filter for Streams",
   "options_ui": {
     "page": "options.html",
-    "open_in_tab": true
+    "open_in_tab": false
```
r2dev2, line 5: does this need to be set to false? would not want to change existing behaviour

src/ts/content_scripts/injector.ts
```diff
@@ -10,6 +10,27 @@ for (const eventName of ['visibilitychange', 'webkitvisibilitychange', 'blur'])
...
+    if (hostname === 'holodex.net' && !holodexEnabled.get()) { return; }
+  } catch (e) {
+    if (e instanceof Error) {
+      holodexEnabled.subscribe($holodexEnabled => {
+        console.log( document.referrer );
+        console.log({ $holodexEnabled });
+      });
+      console.debug('Could not get hostname', { e });
+    } else {
+      throw e;
```
r2dev2: Can you remove the `throw e`? I would not want the function to crash if something goes wrong.

Also, can you replace the debugging of checking `holodexEnabled` with that singular `console.debug`?

r2dev2 (changes requested):
Sorry for taking so long to review this, was a bit busy with university. I have put a few pieces of feedback.

Also, this week, I'm going to be working on the mv3 re-write so I will be a lot more active on GitHub.

src/manifest.json
XangelMusic: This allows an Options Tab in the Manage Extensions page in Firefox, or just gives a simple pop-up in Chrome.

Its just more cleaner imo, and doesn't change the overall behavior of the extension.

r2dev2: Oh wow I actually did not know that is what that does. That's great! Thanks for adding this.

r2dev2 (approved):

PR #456, opened 2024-04-25
WIP: Fix April 2024 E2E Test Issues

r2dev2:
Needed to do below mods because selenium gets more quirky as time passes

* set explicit window size (`open_embed` doesn't work unless `Embed TL` button is visible)
* close ltl splash screen (`test_embed_resize` can't actually resize the embed pane unless testing tab has focus)


KentoNishi:
closing since #488 covers the current mv3 e2e fix now

PR #459, opened 2024-06-16
Remove Android logic

arkon:
Since I no longer use the extension in the Android app, having this logic around clutters the codebase.


KentoNishi:
can u target the diff to https://github.com/LiveTL/LiveTL/tree/develop instead?

arkon:
Updated.

PR #460, opened 2024-06-16
Make extension action icon open options instead of LiveTL website

arkon:


KentoNishi:
same with this one, can u target the diff to https://github.com/LiveTL/LiveTL/tree/develop

arkon:
I could swear someone told me to target the MV3 branch instead back when I first did this, but done.

PR #461, opened 2024-08-01
Fix: emoji not rendered in translation screenshots

frackfrickle:
Emojis were being rendered as an `img` with a src tag which presumably made the emojis end up not being spotted by html2canvas. This included channel/member-specific emotes too.

Enabling allowTaint and useCORS should allow them to be captured.
src: (https://html2canvas.hertzen.com/configuration)


r2dev2 (approved):
Nice fix! LGTM

exported translation

KentoNishi (approved):

Issue #463, opened 2024-10-17
[Bug] Not picking up mod/verified chat messages on YouTube/Twitch


r2dev2:
On YouTube/Twitch, mod/verified messages are not being picked up.

## Twitch

On [northernlion](https://www.twitch.tv/northernlion), nightbot is not being picked up despite being mod.
![image](https://github.com/user-attachments/assets/55478b0e-9cc5-41cb-bc65-c3c2b646f597)

## YouTube

TODO: update description with a concrete timestamp/testcase, but I have verified that mod/verified messages were not caught.

Update: saw a verified message caught on faunwell stream

Reported by `.polardude` on ltlcord

PR #464, opened 2024-11-25
Fix Twitch Parser

r2dev2:
Twitch updated their message representation in the HTML by nesting the fragments within a div. This caused an issue in our previous parser which only looked at direct children. 

PR changes parser to recursively visit children up to a small depth (3) to search for fragments.

Partially addresses #463 (only the twitch part)


KentoNishi (approved):

KentoNishi (changes requested):
fix eslint things, ill do this

KentoNishi (approved):
nah fck it

PR #490, opened 2026-07-20
Fix embedded TLs being lost when toggling YouTube fullscreen

### LINKED ISSUE #454: [Bug] YT De-loads chat with embed tls while fullscreening
As of ~1.5 months ago, when chat is fullscreened in youtube, chat is unloaded. This has the undesireable side effect of all embed tls being lost when full-screening and exiting full-screen mode.

Reported by lunarhawk on ltlcord

NOTE: am working on mv3 migration right now, it would be amazing if someone else could contribute a fix to this in the meantime!

hoangchristopher:
Restores the embedded translation view after YouTube recreates its chat iframe during a fullscreen toggle. Persists the embed flag in `sessionStorage`, scoped by the chat URL's video ID or continuation token.

Carried the existing proposal into `apps/LiveTL/src/ts/content_scripts/injector.ts` on monorepo `main`, preserving the original author, commits, review thread, and non-draft state. The original author reported Firefox fullscreen verification before the migration; this adapted port has not been browser-validated.

**Recommendation: hold; treat as draft work before merging.** The outstanding review asks for extension-backed storage and deliberate live-video versus replay identity scoping. The carried implementation still uses `sessionStorage`, so that review is not resolved. Recheck fullscreen restoration and switching streams after addressing it.

Fixes #454 once the outstanding review is addressed.

Validation on the ported branch:

- `npm ci`
- `npm run format:check` and `npm run lint:check`
- `node --test .github/scripts/prepare-release.test.mjs` — 10 passed
- `npm run test -- -- --runInBand` — LiveTL: 10 suites, 78 tests passed; the other workspaces currently have placeholder test scripts
- `VERSION=0.0.0 npm run build -- --concurrency=2` — full typecheck, build, verification, and packaging matrix passed

No browser or live YouTube validation was run for this port.

### DIFF
--- a/src/ts/content_scripts/injector.ts
+++ b/src/ts/content_scripts/injector.ts
@@ -9,6 +9,15 @@ for (const eventName of ['visibilitychange', 'webkitvisibilitychange', 'blur']) 
   window.addEventListener(eventName, e => e.stopImmediatePropagation(), true);
 }
 
+const EMBED_STORAGE_KEY = 'livetl-chat-embedded';
+
+// `v` identifies a live chat, `continuation` identifies a replay chat. Scoping
+// the persisted embed flag to this keeps it from leaking onto the next
+// stream's chat frame within the same tab.
+const chatId = new URLSearchParams(location.search).get('v') ??
+  new URLSearchParams(location.search).get('continuation');
+const embedStorageKey = chatId != null ? `${EMBED_STORAGE_KEY}:${chatId}` : null;
+
 async function loaded(): Promise<void> {
   try {
     let hostname = '';
@@ -73,6 +82,7 @@ async function loaded(): Promise<void> {
     );
   };
   const embedTLs = (): void => {
+    if (embedStorageKey != null) sessionStorage.setItem(embedStorageKey, 'true');
     const embeddedParams = constructParams();
     embeddedParams.set('embedded', 'true');
     embeddedParams.set('tabid', frameInfo.tabId.toString());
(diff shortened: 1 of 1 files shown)


KentoNishi:
mm im not sure about this implementation, because the continuation and video id are not always consistently available (ie sometimes we only have the video id, sometimes we only have the continuation). also, we don't use sessionStorage because this is an extension, if we were to do this properly we would need to hook it into an extension persistent store

KentoNishi:
@r2dev2 thoughts?

hoangchristopher:
> mm im not sure about this implementation, because the continuation and video id are not always consistently available (ie sometimes we only have the video id, sometimes we only have the continuation). also, we don't use sessionStorage because this is an extension, if we were to do this properly we would need to hook it into an extension persistent store

Makes sense, I'll switch to using LookupStore since it's also used in store.js. Additionally, I've been able to verify that continuation is present across the enable -> fullscreen -> default view loop, but am wondering if you'd be able to point me to a scenario where it'd be unavailable so I can look into it, thank you!

KentoNishi:
yea there are cases for both in our own embed code. live chat uses `live_chat?v=...`, replay uses `live_chat_replay?continuation=...`, so depending on the mode we may only have one of those. try going to a live livestream vs an archived livestream and inspecting the chat url, theyre diff. fyi there's no way to go from one to the other, ie no (known) method of computing/retrieving continuation tokens from the video id or finding the video id corresponding to a continuation token

PR #501, opened 2026-08-08
Consolidate with HyperChat as monorepo

arkon:
Disclaimer: this is mostly executed by Codex

---

Goals:
- Converts this repo into a monorepo that houses both LiveTL and HyperChat sources + releases, with the intention of removing the need to coordinate changes between the two in separate repos

Out of scope:
- Unnecessary dependency upgrades (but may include some, in order to align some things)
- Simplifying in any way how common/shared logic is authored/consumed across the two extensions
- Bringing in any _other_ common things, like iframe-translator

Some side notes:
- Includes HyperChat history for some better context in the future without needing to refer back to the other repo (although it makes this PR seem much bigger than it is with regards to commits, disregarding the "size" of importing the HyperChat source)


arkon:
(Ended up splitting this up)

